PR #12, opened 2015-04-06
Stuff x4

AliiTea:
Flashes can now be overcharged

Bible now blesses water in beakers when placed inside

New Hairstyles

Changeloggy shit


AliiTea:
The error isn't actually an error.. ignore it

Fox-McCloud:
Just so you guys know--be careful what you port from us; we're AGPLv3--which mandates open source--you guys are GPL3, as far as I know, which could lead to conflict and trouble for you guys--for example, since you ported that flash, that code MUST be made available to anyone who requests it if you guys ever decide to go closed source again.

May want to report this to some higher ups and consider switching to an AGPLv3 license (which binds you to never be closed source ever again) to avoid any potential issue.

TheDZD:
Just to add to what Fox said, this also applies to newer /tg/station and Baystation code, as they have also switched over to AGPLv3.

AliiTea:
we (I) know, we switched to opensource so we could have cool shit from other places

Also I am a headcoder for HippieStation and I think we already switched to AGPLv3 but it was on our bugged repository so its just a matter of throwing the license on this copy

Fox-McCloud:
@Memendia 

Sounds great!

PR #58, opened 2015-04-12
Adds shadowling antaghuds

Crystalwarrior:
Adds shadowling antaghuds
Fixes shadowlings not being greeted or initialized properly when antagged manually


Chronitonity:
wow this was done fast oke

PR #60, opened 2015-04-12
Added one more wrassling belt for vault, added two wrassling belts in 

Crystalwarrior:
Box gloves lockers for new ring now contain wrassling belts, as well as one additional belt in vault. Woo

USED MAPMERGE.


j-mie:
Yes you did Mapmerge correct

PR #83, opened 2015-04-17
Poolupdate - by Kokojo and Crystalwarrior

Crystalwarrior:
This will have quite a few conflicts, so that'll need fixing.


WhiteFox08:
see #84

PR #86, opened 2015-04-18
Various tweaks and fixecs

Crystalwarrior:
Check commits


JohnGinnane:
There's some improvents I have in mind but I'll make them on a while when I have keyboard

PR #98, opened 2015-04-21
Adds plasma fist to wizman so he can finally use it

Crystalwarrior:
Plasma fist is great for those who want some added challenge while slightly broadened skillset.
Plasma Fist is super difficult to use, especially if you're swarmed by people with tasers.


AliiTea:
+1

PR #114, opened 2015-05-01
Master

Weisyl:
Adds the AK-922 battle rifle, found in the floor safe of the Russian Station, a bunch of Russian-themed gear, also on the Russian station. The AK-922 also spawns via summon guns.


SplergeBucket:
uh

PR #117, opened 2015-05-05
Buster

HippieStation13:
Copy+paste from original commit before issue with git occured

"Use buster.txt to add IP addresses for public VPN and Proxy IP
addresses. One line per IP.

Alerts admins and logs into the serverlogs. Does not alert users that
they are using a known one, only admins.  This does not ban them, as
someone may legitly be using a VPS/Proxy

At most this just points out people using it. Much like if you have two
people in the same house on two different computers playing.  Only
useful for people using said IPs, but may or may not be useful for
someone running a virtualbox and grabbing a random public IP for sites.
Consider this more of an early warning system. Though it has it flaws,
only IPs entered will be ones it checks. More robust system to come.

Note: THIS DOES NOT BAN THEM."

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -39,6 +39,7 @@
 	var/Ticklag = 0.9
 	var/Tickcomp = 0
 	var/allow_holidays = 0				//toggles whether holiday-specific content should be used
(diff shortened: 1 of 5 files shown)


SplergeBucket:
scooter, you a straight ace busta

PR #120, opened 2015-05-09
fixes blowout

Redswandir:
the rad storm event is FINALLY done

it's true to my vision of whatever the fuck it was like a year ago

i don't think I've gone a bit far in any place tho

and i mean flashing lights is the key to all of this

cause it's a better atmosphere than we've ever had before

and it's sorta like poetry, they rhyme

every station kinda rhymes with the last one

hopefully it'll work


Redswandir:
uh oh god what

Redswandir:
no

PR #122, opened 2015-05-10
rip laggy rsa

Redswandir:
cant be arsed to fix it atm got fucking classes to do


HoaxIT:
Do not merge, i'll fix it.

PR #140, opened 2015-05-22
Fixes borgs repairing blast doors

collinxchu:
Because rglass is different from rglass/borg

### DIFF
--- a/code/game/machinery/doors/glassblastdoor.dm
+++ b/code/game/machinery/doors/glassblastdoor.dm
@@ -123,6 +123,16 @@
 					playsound(loc, 'sound/items/Deconstruct.ogg', 40, 1)
 				else
 					user << "<span class='notice'>You need more glass.</span>"
+		else if(istype(W,/obj/item/stack/sheet/rglass/cyborg))
+			var/obj/item/stack/sheet/rglass/cyborg/GL = W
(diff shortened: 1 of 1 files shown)


ghost:
PR Looks great, but has some duplicate code, how does this look?

```
        else if(istype(W,/obj/item/stack/sheet/rglass) || istype(W,/obj/item/stack/sheet/rglass/cyborg))
            if(istype(W,/obj/item/stack/sheet/rglass))
                var/obj/item/stack/sheet/rglass/GL = W
            else if(istype(W,/obj/item/stack/sheet/rglass/cyborg))
                var/obj/item/stack/sheet/rglass/cyborg/GL = W
            if(health <= 0)
                if(GL.use(2))
                    user << "<span class='notice'>You repair [src].</span>"
                    health = 20
                    update_health()
                    playsound(loc, 'sound/items/Deconstruct.ogg', 40, 1)
                else
                    user << "<span class='notice'>You need more glass.</span>"
```

collinxchu:
I don't believe that's possible, what with having to define
var/obj/item/stack/sheet/rglass/cyborg/GL = W

It's sloppy, I know, but I only dabble in byond code, not my forte.

collinxchu:
Looks good, and prettier. Done.

PR #149, opened 2015-05-28
Goodbye belts

Kokojos:
This removes the belts from the ring's locker.
Also add a second engingun to engineering so HOPEFULLY the SM can be used properly.


lcass:
Yer , I support this one , those belts are annoying as hell and were not goon. I mean even then they had it as a traitor item.

SplergeBucket:
:+1: happy to see how it'll go

PR #164, opened 2015-05-30
[DNM] - Paradise trams.

ghost:
Ports https://github.com/ParadiseSS13/Paradise/pull/1106


lcass:
, ill implement this somehow into the game. We would definately need some custom floor times so it doesnt look like shit.

lcass:
hmm , some work may need to be done here. The trams are ridiculously slow , like so slow you might aswell make 8 round trips in the same time it would take the tram. Secondly they dont rotate at corners which is good it you want pods in space however if you actually want anything that will fit down a hallway but will also be useful you just cant. Maybe a joint and some carriges? Ill experiment ive had some ideas.

PR #174, opened 2015-06-03
Duct tape fix

lcass:
Stops people ductaping you from a mile away. Whoever put it into afterattack SHAME ON YOU.


ghost:
funfact, attack() also triggers via security cams.

PR #186, opened 2015-06-07
Show Game Mode Odds in the getrev command.

ghost:
Show Game Mode Odds in the getrev command.


SplergeBucket:
this is an ehhhh thing
probably wont cause any harm tho

PR #188, opened 2015-06-08
HTML interface module + playing cards using the new module

nullquery:
_Ported from Baystation12 upon request of @HoaxIT_
Original PR @ https://github.com/Baystation12/Baystation12/pull/9732

This PR introduces a new method of dealing with HTML browser windows. My intention is to replace every component in the game that uses browse() with this module.

The module is documented in the main file and there's an example in the form of playing cards which have been modified to use the new system.

There are a lot of improvements to this system over the older systems, notably that it remains easy-to-use yet refreshes parts of the screen without flicker. Unlike NanoUI you specify the parts to replace directly in the source-code, there is no need to tinker with templates.

The HTML windows opened contain the latest stable/legacy version of jQuery for ease-of-use. It's possible to send Javascript calls directly using a function or you can just define something yourself through HTML.

This includes 3 possible versions of the window. The basic template is in /datum/html_interface and consists of a simple window with a browser component. /datum/html_interface/nanotrasen is the NanoUI-lookalike with a custom titlebar. I have yet to acquire icons for the minus and close buttons but I'm sure someone else can oblige in this regard (this component is not yet in use, by the time it becomes in use I hope to have those sprites).

Finally there's /datum/html_interface/cards for the playing cards. They need a subtype because the playing cards have their own resources which need to be preloaded by the client.

Future contributions to the module will be a generic way of authenticating ID cards, so that the computers on the station can be ported to make use of the new system.

Ideally NanoUI becomes deprecated, I'm not sure what the feeling about that is but I think it's necessary to keep the system as simple as possible.

image


ghost:
You're amazing, pulling as soon as Travis goes green.

nullquery:
**To anyone looking to develop this further:** Please contact me on Github or at the very least develop from [my own repository](https://github.com/nullstation/nullstation) and then push it to your own codebase (or I'll do it for you as you've just witnessed).

HoaxIT:
@nullquery merged, thanks.

Issue #62, opened 2015-04-13
mussle doesnt stop screaming


Kokojos:
Self-explanatory

Chronitonity:
mussle

Crystalwarrior:
mussle?

WhiteFox08:
mussle!

AliiTea:
Kokojo can you explain

Kokojos:
zzzzzzzzzzzzzzzzzzzzzzzzzzzzzz

Crystalwarrior:
Fixed nuzzles allowing you to scream on Atlas station. Also fixed silence variable not having the same effect. Check my pr's on atlas, currently on phone.

AliiTea:
Then close the damn issue you cucklord D:

ghost:
This is not atlas, still not fixed.

AliiTea:
Then maybe you should get off ur ass and take our fix for it D:

PR #212, opened 2015-06-14
Fix to issue #62, screaming while muzzled.

Jaygriff:
Thanks to Crystal at atlas for the fix.


Jaygriff:
Tested compilation and confirmed working

Issue #215, opened 2015-06-17
You can put poly acid in a spray bottle by using a syringe


killer653:
You can bypass being unable to put poly acid into a spray bottle by using a syringe.

How to reproduce:
1. Get a spray bottle
2. Make poly acid
3. Attempt to put poly acid in bottle via beaker, see it doesn't work
4. Take the poly acid out via syringe and inject it into the spray bottle
5. have fun with your killing weapon

ghost:
known bug.

AliiTea:
Easy fix

Nokington:
This was fixed

PR #232, opened 2015-06-29
Disabled injecting chemicals directly into spray bottles.

Jaygriff:
Could be used later with a checker to more specifically filter out banned chems. Quick fix for now

issue #215


killer653:
Does this prevent droppers from being used, too?

Jaygriff:
Nope, will pr that fix tomorrow since I forgot.

ghost:
bad way to fix this.

PR #234, opened 2015-06-30
Poly acid fix for syringes and droppers

Jaygriff:
Hopefully thats all of them.
#215

### DIFF
--- a/code/modules/reagents/reagent_containers/glass.dm
+++ b/code/modules/reagents/reagent_containers/glass.dm
@@ -11,6 +11,7 @@
 	volume = 50
 	flags = OPENCONTAINER
 	can_examine_reagents = 1
+	banned_reagents = list("pacid","sacid")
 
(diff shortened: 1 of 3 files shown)


code/modules/reagents/reagent_containers/glass.dm
```diff
@@ -11,6 +11,7 @@
 	volume = 50
 	flags = OPENCONTAINER
 	can_examine_reagents = 1
+	banned_reagents = list("pacid","sacid")
```
ghost: this is bad, we still need acids in glass

Jaygriff: This is just a list of the banned reagents. It is only called in the one of the functions and isn't called at all in this glass.dm file.. Infact I can edit that out as it does absolutely nothing.
Thanks for pointing it out

PR #240, opened 2015-07-11
Goodcurity & admin alert levels

nalarmg:
Per Nokingtons request

Allows admins to set any alert level at will via the secrets panel.

Allows those with the goodcurity achievement to set code delta via the communications console. Inline with the goodcurity "fix" (as it was previously existent but lost, apparently), code delta can be changed to code red via the communications console now, allowing goodcurity to revert martial law when it's no longer needed..

### DIFF
--- a/code/game/machinery/computer/communications.dm
+++ b/code/game/machinery/computer/communications.dm
@@ -31,7 +31,7 @@ var/const/CALL_SHUTTLE_REASON_LENGTH = 12
 	var/const/STATE_ALERT_LEVEL = 8
 	var/const/STATE_CONFIRM_LEVEL = 9
 	var/const/STATE_TOGGLE_EMERGENCY = 10
-	
+
 	l_color = "#FFFFFF"
 
 	var/status_display_freq = "1435"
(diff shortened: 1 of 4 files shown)


code/modules/admin/admin.dm
```diff
@@ -483,6 +483,14 @@ var/global/floorIsLava = 0
 			<A href='?src=\ref[src];secretscoder=infinite_sec'>Remove cap on security officers</A><BR>
 			<BR>
 			"}
+	if(check_rights(R_FUN,0) && check_rights(R_DEBUG,0))
```
nalarmg: Oh wow. I can't believe I missed this, actually. Redundant check. Needs to be adjusted.

code/game/machinery/computer/communications.dm
```diff
@@ -420,10 +420,13 @@ var/const/CALL_SHUTTLE_REASON_LENGTH = 12
 		if(STATE_ALERT_LEVEL)
 			dat += "Current alert level: [get_security_level()]<BR>"
 			if(security_level == SEC_LEVEL_DELTA)
-				dat += "<font color='red'><b>The self-destruct mechanism is active. Find a way to deactivate the mechanism to lower the alert level or evacuate.</b></font>"
+				dat += "<A HREF='?src=\ref[src];operation=securitylevel;newalertlevel=[SEC_LEVEL_RED]'>Red</A>"
```
ghost, line 434: No! This breaks two gamemodes, malf & nook ops. 

`The self-destruct mechanism is active. Find a way to deactivate the mechanism to lower the alert level or evacuate.`

besides that, `goodcurity` is not used, since we're not using the achievements.

SplergeBucket:
cant we just run goodcurity like we run the snowflake reward system

ghost:
No, i'd rather have it use the achievement system, and fix that on the box.

PR #247, opened 2015-07-12
stage 6

ghost:
ports https://github.com/tgstation/-tg-station/pull/9847


Nokington:
It needs a nerf I think. Getting set on fire within 25 tiles of it is a bit much.

TDcoolguy300:
Current issue with this port, this is referring to the supermatter_Shard object, and not VG's supermatter object, needs to be fixed at
"if(istype(A, /obj/machinery/power/supermatter_shard) && !consumedSupermatter)" 
While the supermatter we use is" /obj/machinery/power/supermatter" and the shard subtype is "/obj/machinery/power/supermatter/shard"

ghost:
oopsies.

Nokington:
This still needs a nerf from 25 tiles. Thats just too much for people to random burst into flames

collinxchu:
To be honest at 25 tiles you're being sucked towards it anyway.

I don't think it needs a nerf but the fact that both engines are so close together does make it a concern. Maybe a map edit is in order.

ghost:
Nerfed to 10.

Nokington:
@collinxchu A map edit would be nice yes but it just won't happen because the map is designed to just be a death trap. Why do you think the grav gen is in the middle of both the engines.

ghost:
@Nokington merge ready, pls merge.

Nokington:
@Chnkr I'm passing it by the heads first.

Nokington:
My opinion on this has changed to do not want because this would most likely happen every round. It probably never happens on TG because I believe the supermatter shard is only ordered from cargo. But for us we just have a supermatter on station and odds are that it is gonna happen everytime the singulo is released. I would however be fine with this if we remap engineering so it isn't such a deathtrap.

ghost:
the station is meant to be a fukken deathtrap, the chance of the singulo hitting it, is low since its `rng` based, i'd say we try this out for a week.

AliiTea:
@Chnkr Great idea

code/modules/power/singularity/singularity.dm
```diff
@@ -349,6 +374,14 @@
 	return
 
 
+/obj/singularity/proc/combust_mobs()
+	for(var/mob/living/carbon/C in orange(10, src))
+		C.visible_message("<span class='warning'>[C]'s skin bursts into flame!</span>", \
+						  "<span class='boldannounce'>You feel an inner fire as your skin is suddenly covered in fire!</span>")
```
nalarmg, line 380: This is a bit of an awkward sentence to read. 

"You feel your insides boil as your skin is engulfed in flame!" would be a better alternative.

Nokington:
This isn't getting anywhere and none of the heads want it because it will happen every round until the SM is removed.

PR #248, opened 2015-07-12
this is a death-trap.

ghost:
semi-port of https://github.com/tgstation/-tg-station/pull/10092


Nokington:
And as long as the server can handle it, we don't need more things to cause lag.

PR #250, opened 2015-07-13
Slight debuffing of the throwing knife

lcass:
Lowers the damage when thrown to be the same as that of a thrown null rod


Nokington:
Make it 15 and its good to merge. Just tested 10 and its way too weak now

Carbonhell:
Are you sure 15 is enough?I mean, syndicate cards cost one tc for an entire deck which should be 30-40 cards,each one does 10 brute damage when thrown and has an high chance of embedding,paying 6 tcs only for 5 more brute damage seems pointless.

Nokington:
Who actually buys syndicate cards

Carbonhell:
...Me,i once killed an entire crew with 2 decks

PR #275, opened 2015-07-18
Updated changelog

Weisyl:
This is the updated changelog for my recently merged PR.


Weisyl:
@Nokington the merge pull request is part of this for some reason but I think it's just basically merging this master branch into my fork for it to be fully updated? I did something like that earlier, and it still lists the only file changed as being the changelog, which is what this PR is for.

Issue #252, opened 2015-07-13
Rad storms borked


collinxchu:
The event triggers, no overlay, says the radstorm passed, then it occurs and kills everyone because they thought they were safe.

Only seen it once but yeah.

ghost:
the overlay is removed.

Jaygriff:
the delay on the end of storm announcement is not long enough.

PR #287, opened 2015-07-19
Added delay in rad storm end message

Jaygriff:
per issue #252


ghost:
this is bad.

Nokington:
Why is that

PR #298, opened 2015-07-21
Fixed not being able to transfer banned reagents from beakers to othe…

Jaygriff:
…r bottles

You still cannot transfer it to spray bottles.


Nokington:
So they can transfer banned reagents into bottles? Like into alcohol bottles?

Jaygriff:
Yes.

Issue #299, opened 2015-07-21
Frag12 shotgun round is not explosive as it says it 


Jaygriff:
Title explains.

PR #312, opened 2015-07-23
Added the explosive part to the explosive slug

Jaygriff:
per #299

img


Nokington:
Where do you even get explosive slugs?

Jaygriff:
I dunno man I just work here

Weisyl:
I think that you can make them from techshells, anyway, the desc and name implies they should be explosive, but either the guy who originally added them got lazy or forgot.

Issue #305, opened 2015-07-22
Rpd nerf


Jaygriff:
Causes huge issues

Nokington:
yes please

Kokojos:
This has been fixed thanks to JBN

PR #314, opened 2015-07-23
RPD port and fix.

Jaygriff:
per #305

This update fixes multiple issues with the RPD and, unintentionally, improves (in my opinion) the look of the interface.

rpd

Also prevents the infinite stacking of disposal units and increases the time it takes to actually spawn in the units..

no more spam!
spam

Other background changes that were made is a port of the pipe creation script from TG and some checkers in the atmospheric pipe code itself, nothing was functionally changed.


Kokojos:
Merging

Issue #330, opened 2015-07-25
Disposal pipe in engineering is 1 tile off


Chronitonity:
Should be 1 tile up.
http://i.imgur.com/o0mEMJs.png
Once again found by NoscopeToolbox
Can't even be fixed in game because of pipe welding bug.

Nokington:
This wasn't me either. I swear.

PR #333, opened 2015-07-25
Fixing mapping issues.

Nokington:
Fixes https://github.com/HippieStationCode/HippieStation13/issues/330 and http://www.hippie-station-13.com/forums/viewtopic.php?f=21&t=2051&view=unread#unread


nalarmg:
Code seems fine.

PR #335, opened 2015-07-25
Adds changelog for initial cat fixes

nalarmg:
Forgot changelog for initial cat fixes. Oops.


Nokington:
because of the recent changelog change

PR #338, opened 2015-07-25
G17 and Abzats polishing update.

Weisyl:
This update adds a new set of sprites for the Abzats to make it more distinguishable from the L6 SAW. The G17 is now compatible with suppressors and have an improved sprite for the magazine. Also (possibly) fixes some guns being able to have suppressors attached when it wasn't intended. The G17 is also properly named in the traitor / nuke ops uplink.


WhiteFox08:
Got a pic of them sprites?

Weisyl:
I can get one in a moment.

WhiteFox08:
Also is this going to nerf the universal suppressor as it was intended to be fitted on any gun

Weisyl:
Erm, I'm pretty sure it wasn't. Weapons like the C-90, AK-922, and even the Desert Eagles could have the universal suppressor attached even though they had no sprites for when a suppressor was on them. Also, the reason that was happening was because the parent weapon / pistol would have the can_suppress var set to 1, and everything under would HAVE to have that set to 0 for themself if they didn't want the suppressor on. In that case, it seems more like an oversight.

WhiteFox08:
Very well I would allow the c 90 to be silenced for secret ops.

Weisyl:
C-90 being and AK-922 suppressed would be cool, but that'd probably have to be decided on via a trello card since they're both pretty powerful. Not to mention they'd both need new sprites for the attached suppressor.

WhiteFox08:
Hmm perhaps reduce the brute damage when a gun is  suppressed?

Weisyl:
G17 Mags: http://puu.sh/jcMW4/1142ad07d1.png  http://puu.sh/jcMYn/09b486068c.png
Abzats: http://puu.sh/jcNca/6ba5b647f7.png   http://puu.sh/jcNdi/f803bfc5db.png
G17 Suppressed: http://puu.sh/jcN8K/e637a47261.png   http://puu.sh/jcNaE/31d8c703d5.png

Weisyl:
As for making a suppressed weapon do less damage - you'd have to make it so that when a suppressor is attached, every single projectile fired from that gun then does less force damage, which would probably be fucky, complicated, or cause some unnecessary CPU strain. Also from a more realistic standpoint, that'd only happen if you're using subsonic ammunition, which isn't required for all suppressors, nonetheless a futuristic one.

Nokington:
If this just changes sprites its good. But if it does something else then i'm gonna wait until I talk to @HippieStationCode/headadmins

Weisyl:
@Nokington besides the sprite changes, the G17 is now able to have a suppressor attached, and rather than being named "Syndicate Revolver" in the uplink (a typo), it is now properly named "Glock 17 Handgun." Besides that, the G17 magazine's new sprite allows you to see if it has either 0 rounds or 1 - 17 rounds. Although additionally, the C-90, AK-922, and other weapons such as the Desert Eagle are no longer compatible with suppressors since it didn't seem intentional in the first place, as they had no sprite that showed a suppressor attached. TL;DR a parent weapon with the can_suppress var being set to 1 made it so that every weapon that didn't allow for suppressors to be attached had to have can_suppress = 0 , however this was absent on some weapons.

Nokington:
Alright, quick question though, can you only get the Glock from an uplink?

Issue #302, opened 2015-07-22
Rodgun anchor not being turned off on movement


blobbernaut:
I was hit by a rod gun and pinned to a wall. I used a hand tele to escape, but I was still unable to walk. 

This is an issue because it doesn't make sense.

JustinBNewton:
I suggest making it so it rips whatever is holding the person to the wall out and causes a massive amount of brute damage.

nalarmg:
It makes sense code wise - at the moment there's no trigger for making you able to move again if this happens. The easiest way to deal with this is, if an object is embedded/the person is pinned, make them able to move but keep the object embedded. Fairly trivial fix.

Chronitonity:
Couldnt you add a way to "resist" out of being pinned to a wall? the only issue i could think of is that it would interfere with the resist function used by handcuffs/cablecuffs/ect

Tomeno:
you have to rip the rod out as far as I know

PR #345, opened 2015-07-26
Fixes pinning on forced movement and icon when un-pinned

Tomeno:
Fixes issue #302 and another bug I found out about while fixing this - your offset when pinned doesn't go back when you get unpinned.

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -39,6 +39,13 @@
 		if(M.lying)	return
 		if(istype(M, /mob/living/carbon/human))
 			var/mob/living/carbon/human/H = M
+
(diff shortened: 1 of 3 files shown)


Tomeno:
The second commit's diff is a little messed up in Github because I didn't update the branch before editing stuff so it lists all the other things as a change even though they're not changed from the current version on here...

Kokojos:
:+1:

Nokington:
Please update your branch so we don't have to revert this if one of those other things get reverted.

Tomeno:
I have updated my fork, but it still shows up like this for some reason :( Can't do anything about it

Nokington:
aaaaaaaaaaaaaaaaaaaa
fine
i'm blaming you for all my problems now

PR #354, opened 2015-07-27
Mutagen added to biogenerator

Jaygriff:
For when/if it is accepted on trello. Feature.


Chronitonity:
Why not just keep it to chemistry only and promote Chemistry doing more than just making lube and polyacid? Seems like a needless addition and removes jobs working together

Nokington:
Its pretty expensive to get. He didn't post a picture here but on trello he did. It costs 200 points or whatever its called in the bio-gen to get one bottle of it. While the E-Z harvest costs 10 points to get.

Chronitonity:
Alright, I still think its a needless addition but that isn't quite as bad as I originally thought

Jaygriff:
I want this to be a alternative to breaking into chemistry to steal it if the chemists aren't cooperating and make the botanists jobs more worthwhile if that is the case. Plus it is only 50 units and it can take a whole lot more to get anything out of the RNG at times.

Nokington:
@JustinBNewton How much mutagen does the bio-gen give?

Jaygriff:
50 units

collinxchu:
Oftentimes the chemists are either absent entirely or refuse to give out mutagen because it's too dangerous. I like the idea that botanists can earn their own progression rather than forcing them to break into chem or rely on the kindness of another spessman.

Nokington:
Approved

PR #360, opened 2015-07-27
Fixes the changelog

Tomeno:
I'm sorry


Tomeno:
i have no fucking clue why it lists those 2 commits when they have nothing to do with this but okay whatever

also it has 100+ contributors what the flying fuck

Issue #353, opened 2015-07-27
Adamantine golems are affected by syringes.


blobbernaut:
They're made of adamantine. They shouldn't be affected by syringe guns. Thickmaterial pls

JustinBNewton:
They shouldn't be affected by syringes. Even hard suits prevent needles... I feel like a pure metal humanoid would be able to deflect that.

Tomeno:
Workin' on it

PR #361, opened 2015-07-27
Makes golems not injectable

Tomeno:
Addresses issue #353

Also those other 2 commits are still fucking listed as changes jesus christ people fix the perms so I can make a branch on the main fork


Nokington:
@Tomeno Please apply for coder on the forums so you can be a part of our coder chat.

Issue #359, opened 2015-07-27
Mass drivering something puts it on the other side of the station


Chronitonity:
Dunno if random z levels in space are enabled, but if you mass driver something it lands near arrivals. 

This is kind of a huge issue seeing how you mass driver things out of the chapel you expect them to be somewhere in deep space, not on the other side of the station.

Nokington:
Pretty sure you just have bad luck.

Kokojos:
Bad luck, approx 1/4 chance

Chronitonity:
Happened 100% of the time I tested it. Tested it multiple times in multiple rounds with @TDcoolguy300

Chronitonity:
Also tested with Nalar, happens literally 100% of the time. If you try and go back to the mass driver and push off to the left of the station, you always end up on the asteroid.

nalarmg:
Yep, tested and confirmed. This is a consistent thing. Legitimate issue.

Kokojos:
Ah, I think I know why. It's when I changed the size of the z-levels - I moved some things around. I think it's z-level 3 that needs something to stop the mass-driver from looping around.

Kokojos:
Actually I tested it again today and your 100% chance dosen't match with my result. Still, il ''nerf'' the rate of returns.

Issue #356, opened 2015-07-27
mapping issues


ghost:
all arcade machines, shuttle console on syndie ship in asteroid belt z level.
cant be moved/deconned

Kokojos:
Tested, can't reproduce.
First of all there's no arcade on the asteroid belt z-level
And I was able to deconstruct all the things in the z-level, but not move them because there's no gravity.

Gonna close unless you have specifications.

ghost:
i'm talking about the solar controller console on the asteroid belt station, if you screwdriver it you get no message and it then dissapears if you wrench it

Issue #351, opened 2015-07-27
Centcomm is not up to code


Nebyab:
There's a hole in the floor between the magic mirror chamber and the security dock in Centcomm.

http://i.imgur.com/ZxQEl4c.png

pls fix

Nokington:
this is so minor that it will probably never be fixed.

Nokington:
Or @Kokojos could just be a cuck and fix it quickly.

Kokojos:
Centcom needs some love.

Weisyl:
I can fix this in a bit with a polishing update.

Nokington:
It was already fixed in koko's PR

Weisyl:
This should be closed then.

Nokington:
It hasnt been merged, thats why it isnt closed

Kokojos:
And the new commit fucked it up. Il have to do it later.

PR #378, opened 2015-07-28
Normalise files from the new mapmerge tool.

Kokojos:
- fix a few map bug, including
  https://github.com/HippieStationCode/HippieStation13/issues/351
  https://github.com/HippieStationCode/HippieStation13/issues/359

And maybe https://github.com/HippieStationCode/HippieStation13/issues/356


Kokojos:
I say ''Normalise'' because it adds a line in each file. Ergo those should be new ''base'' files.

Jaygriff:
Did it work?

PR #380, opened 2015-07-29
Update game_options.txt

WhiteFox08:
Takes Abduction out of the Rotation until I can deploy fixes.


nalarmg:
It needs mentioning that these changes do not take effect on the actual server unless me or someone else with box access updates it. Please inform someone who does have access when the config needs to be changed to accommodate stuff like this.

Issue #379, opened 2015-07-28
You can morph nukeops' mask into the 3 gasmask variants


Carbonhell:
As the title says. It shouldn't be a thing, i think.

Jaygriff:
Its a inheritance issue

PR #390, opened 2015-07-30
Fixed Other Variants of the gas masks being changeable

Jaygriff:
per  #379


Nokington:
@JustinBNewton This could use a changelog but it doesn't matter

Issue #393, opened 2015-07-31
Blowing the lighting circuit on APCs as AI drains power quickly


Jaygriff:
Doesn't make sense that you can do it more than one time after the lights are blown.

Kokojos:
IMO it would need an option for the AI to actually drain the power.

Jaygriff:
I'd be down for that but the blowing the lights over and over to drain the power is just stupid and gimmick y

PR #395, opened 2015-07-31
Overload doesn't drain if it didn't blow up any lights

Tomeno:
Fixes issue #393 

Changelog causes conflict :(


Nokington:
@Tomeno Make a changelog please

PR #396, opened 2015-08-01
Fixes Walking Mushrooms attacking people, adds the bite effect to their attack sound

TDcoolguy300:
Does as the title says, Walking Mushrooms will no longer attack people, and adds an attack sound for them to use. Now we can have Mushroom fighting rings again, without everyone outside dying


Jaygriff:
Good

Kokojos:
Saw the test in game, this is good.

Nokington:
Changelog please

PR #411, opened 2015-08-09
Turrets are cooler now.

ninjacha:
http://prntscr.com/82n12o
3kpkbg3


ninjacha:
Oh, woops, I commited 28b11ea to turret-port instead of another branch.

TDcoolguy300:
We can't auto-merge this due to it not being a bug fix, i'll try and bring it up with the headmins to try and approve this for merging

lcass:
revert the wrassling belts and entirely remove the defines file , don't mess with that. Other than that it looks good.

ninjacha:
I'm now permabanned for scarring 2 tiles now, so no thanks.

Issue #408, opened 2015-08-06
Supermatter respawn bug


TDcoolguy300:
If you were to throw a locker which contains a person into the Supermatter, the person's body AND Ghost will delete, causing their client to crash, allowing them to rejoin as a new character. I have done a test with Kruor to confirm this

WhiteFox08:
Yea, SM doesn't check for occupants and just deletes it along with the locker.

Jaygriff:
Fixed

PR #412, opened 2015-08-10
Supermatter fixes

Jaygriff:
per #408


tgstation.dme
```diff
@@ -6,6 +6,80 @@
...
+#define FILE_DIR "sound/violin"
+#define FILE_DIR "sound/voice"
+#define FILE_DIR "sound/voice/complionator"
+#define FILE_DIR "sound/voice/complionator/bees"
+#define FILE_DIR "sound/weapons"
+#define FILE_DIR "sound/weapons/effects"
+#define FILE_DIR "sound/welcome"
+#define FILE_DIR "tools"
+#define FILE_DIR "tools/AddToChangelog"
+#define FILE_DIR "tools/AddToChangelog/AddToChangelog"
```
Nokington: RIP

Nokington:
So this changes the supermatter into the shard?

SplergeBucket:
What actual change was made to the supermatter being something you can drag around and murder the station? Is this just removing it and changing it to TG's stats.

Chronitonity:
Does this fix anything

Jaygriff:
#408

tgstation.dme
Jaygriff: Yes RIP

Chronitonity:
So it fixes that by just porting TG's version and then (if im reading this correctly) removing it from the map?

Jaygriff:
Negative. I ported TGs version and made it into the base supermatter, as they only have the shard. Im looking at something right now, DNM.

tgstation.dme
Nokington: Go into preferences and turn this off to remove it

Jaygriff: Preferences on what?

Nokington: Hang on, I gotta remember where. Are you using sourcetree

Jaygriff:
All good @Nokington

Nokington:
@JustinBNewton I will have to talk to the other heads, because I am not sure about replacing the SM with the shard.

Jaygriff:
@Nokington Thats what TG did. However I modified it back to the SM with their back end code to fix the issues we had with it.

IE nothing was changed but the background stuff that no one can see and it still behaves the same way but doesn't cause re spawns and functions just a little bit better doing checks.

WhiteFox08:
:+1:

code/modules/power/supermatter/supermatter.dm
```diff
@@ -230,172 +173,130 @@
...
+/obj/machinery/power/supermatter/singularity_act()
+	var/gain = 100
+	investigate_log("Supermatter shard consumed by singularity.","singulo")
+	message_admins("Singularity has consumed a supermatter shard and can now become stage six.")
+	visible_message("<span class='userdanger'>[src] is consumed by the singularity!</span>")
+	for(var/mob/M in mob_list)
+		M << 'sound/effects/supermatter.ogg' //everyone goan know bout this
+		M << "<span class='boldannounce'>A horrible screeching fills your ears, and a wave of dread washes over you...</span>"
+	qdel(src)
+	return(gain)
```
TDcoolguy300, line 224: We do not have the Stage 6 ported, so this really shouldn't be in here, although this is just the messaging bit, and should not cause any runtimes if left in. Otherwise, this looks good to me

code/modules/power/supermatter/supermatter.dm
```diff
@@ -27,126 +14,71 @@
...
 
 
-#define WARNING_DELAY 30			//seconds between warnings.
+#define WARNING_DELAY 30 		//seconds between warnings.
 
 /obj/machinery/power/supermatter
-	name = "Supermatter"
-	desc = "A strangely translucent and iridescent crystal. \red You get headaches just from looking at it."
+	name = "supermatter shard"
+	desc = "A strangely translucent and iridescent crystal that looks like it used to be part of a larger structure. <span class='danger'>You get headaches just from looking at it.</span>"
```
TDcoolguy300: Not changing the sprite, so the previous description should be used and not the new one, also needs it's name changed.

Jaygriff: Whoops, could have sworn i diff checked that in, my bad!

lcass:
I am currently coming up with a modification to the way we use the supermatter , if it functions in the same way that the full size supermatter does (produce plasma and heat etc) then I'm fine with the change.

Jaygriff:
@lcass Functionally the same minus the delete from round side affects.

lcass:
Alright , if the heads agree then this can get merged.

Jaygriff:
Dont think it needs a vote. Its a bugfix not a new feature

lcass:
yes but it prevents antag usage which is a balance change.

Jaygriff:
Antag usage? What are you talking about?

lcass:
draggy draggy rad death otherise known as the erik hanson effect

Jaygriff:
I dint realise that was a festure specific to hippie, but i shall add it in tomorrow. My mistake.

lcass:
oh no it was something they were discussing removing. That's why there was a vote.

WhiteFox08:
Pretty sure heads wanted it not to be movable.

danhibiki:
Its not that it would be totally unmovable. The idea is that we would have a shard containment cart, but by default a player should not be able to drag it.

So yes, unmovable as of now by dragging via mob

SplergeBucket:
I wanted it to dust people who tried to pull it.

Nokington:
Unmovable and dusting people to try to pull it sounds good to me

ghost:
This change doesnt actually do anything, it just removes a whole lot of features from it, it will no longer pull in tiles, and object when it asplodes.

The explosion itself is also made smaller, no idea why.. why? @JustinBNewton it seems that you are just changing random variables.

Jaygriff:
Please check issue 408 before commenting beyond this point.

Jaygriff:
If youd like to cook up your own fix, feel free.

ghost:
@JustinBNewton i dont see how that will fix this issue? You can easily check if the locker has any occupants by looping trough `contents` or calling `dump_contents` before its qdelled.

Jaygriff:
Its the way TG deals with it. Were tg based. Its the way we'll deal with it. It was talked over with the team and decided this way was the best.

ghost:
The issue is still there, even with this change.

Chronitonity:
If it stops dragging tiles and objects then this doesnt seem like a pure fix, just like a change of how a feature works.

Jaygriff:
@chnkr its clear you did not test it for yourself. It fixes the issue.
Chron, tiles getting pulled in works exactly how ot used to. If it doesnt, ill fix it as soon as i get a final word on dragging.

PR #460, opened 2015-08-23
Puts Diseases and Object controllers back on updatequeue

WhiteFox08:


ghost:
The `updateQueueInstance` is slow, and is removed in the latest version, using a normal loop is faster.

PR #502, opened 2015-08-31
Chem balancing

lcass:
Removes ethyl alcohol , replaces it with the ethanol consumable. Buffs bearium to be easier to use , debuffs diethyl zinc explosion , fixes tabun recipe , buffs lots of other difficult chem recipes and their uses and make N-amino azidotetrazole more difficult to produce.


lcass:
Also fixes a bug with the distiller.

PR #507, opened 2015-09-01
chloral emergency fix

lcass:
Fixes chloral hydrate insta killing people , my god what a buggy function reaction_mob is.


nalarmg:
Code's fine

PR #514, opened 2015-09-02
Warning on compile fix

nalarmg:
Fixes awful white spacing between procs & definitions.
Fixes warning on compile.
Mineral floors now set life.


lcass:
Good god , my mistake 0_0. Code is fine , this will cause runtimes so merge this ASAP.

Issue #508, opened 2015-09-01
Electrified doors don't shock anymore.


Carbonhell:
As the title says, if you hack a door or run into an electrified doors, the sparks appear but you don't get stunned nor harmed, therefore any kind of insulated gloves is now useless.

whitefox08:
![xkxb9q5](https://cloud.githubusercontent.com/assets/2702850/9660205/0b658124-521a-11e5-8fcc-39e6aad1ea5a.jpg)
![ft4wr6y](https://cloud.githubusercontent.com/assets/2702850/9660204/0b610da6-521a-11e5-82a9-2085af829ac8.jpg)

Works both for hacking and AI electrified. Are you sure the door had power?

WhiteFox08:
Fixed in #523

WhiteFox08:
Issue is still present relooking into this.

AliiTea:
You don't get shocked from anything.. not just doors

ghost:
rofl 10/10 coding

nalarmg:
You still get shocked by wires if you attack them with objects.

Jaygriff:
Yep, shocking by hitting on wires still works.

Chronitonity:
APCs and Grilles still shock too

PR #523, opened 2015-09-03
Bug Squashing

### LINKED ISSUE #501: Gold slimes extract can spawn Ak-922 wielding russian.
As the title says,i don't think it should be able to spawn human enemies like syndicate or soviets. And it could be used to make an ak-922 farm..oh jesus.

WhiteFox08:
Fixes #501 #508

@lcass I had to pull upstream you probably want to check Chemistry-Recipes.dm in this PR


nalarmg:
Code's okay.

PR #534, opened 2015-09-07
Waste Disposals "Fix"

Pyko1:
Made the Waste Disposals once again shoot the thrash out of Z level instead of the singularity cage wall.
image


Chronitonity:
Honestly my main gripe is that the conver belt is just super stretched and that the room where everything is in is now just huge and open with nothing else in it besides a lever

Pyko1:
I guess i could add some shit or move things around a bit as well.

Pyko1:
Update Image:
capture

Chronitonity:
:+1:

Kokojos:
Too many warning signs.
Also IIRC there's no need for double launchers, one should do.

Pyko1:
I kinda agree on the signs.But with just 1 launcher it will break after a while once the tile under the blast door is depressurised. But placing the vent under the blastdoor might fix tho.

I'll see what I can do.

Pyko1:
Better?
capture

Kokojos:
Just make sure all your signs are correctly labelled and I see no problem with this at first view.

Pyko1:
Yeah, changed the name and desc on those sign.

JustinBNewton:
remove the walls that are in space as well as the catwalks outside the door and leading off the southern catwalk to the north. That area in south arrivals is supposed to be a docking port for larger ships.

img

Pyko1:
As it is now.
capture

Chronitonity:
Catwalk should probably extend more so they almost touch each other.

PR #538, opened 2015-09-08
Fix gravity, borg eating the water, and stop beakers from going in pools

### LINKED ISSUE #286: Water tiles can be deleted by janiborg floor cleaner.
/obj/effect/overlay/water is the actual item.

Kokojos:
fixes #286 and some random bugs no one bottered to check.
also removes the window and warning for headsets and it's no longer necessary.

Also remove all about the beaker functionnality.


Kokojos:
I managed to fail mapmerging. Go koko.

Kokojos:
All good.

PR #557, opened 2015-09-10
Gives security their sechailers back

Chronitonity:
This means that security no longer starts with normal gasmasks in their boxes, but sechailers instead. Someone on TG removed this for some reason.


AliiTea:
@Chronitonity missing the box for the rest of sec
 lines

29
130
180

Chronitonity:
Hold on I'm an idiot

lcass:
Make sure to update the changelog when this is merged if the changelog is updooted.

PR #570, opened 2015-09-12
Updates to the ban system

lcass:
Adds in some extra features to the ban system making it easier to spot evasion attempts and moving the ban detection more server side.

### DIFF
--- a/code/modules/admin/NewBan.dm
+++ b/code/modules/admin/NewBan.dm
@@ -92,11 +92,23 @@ var/savefile/Banlist
 
 	return 1
 
-
+/proc/add_banfile(var/client/c)
+	var/savefile/file = new()
+	file["key"] << c.ckey
+	file["ip"] <<c.address
+	file["cid"] <<c.computer_id
+	c.Export(file)
+/proc/remove_banfile(var/client/c)
+	c.Export()
+/proc/get_banfile(var/client/c)
+	return c.Import()
 /proc/AddBan(ckey, computerid, reason, bannedby, temp, minutes, address)
 
 	var/bantimestamp
-
+	for(var/client/c in clients)
+		if(ckey == c.ckey)
+			add_banfile(c)
+			break
 	if (temp)
 		UpdateTime()
 		bantimestamp = CMinutes + minutes
@@ -122,12 +134,15 @@ var/savefile/Banlist
 /proc/RemoveBan(foldername)
 	var/key
 	var/id
-
+	
 	Banlist.cd = "/base/[foldername]"
 	Banlist["key"] >> key
 	Banlist["id"] >> id
 	Banlist.cd = "/base"
-
+	for(var/client/c in clients)
+		if(key == c.ckey)
+			remove_banfile(c)
+			break
 	if (!Banlist.dir.Remove(foldername)) return 0
 
 	if(!usr)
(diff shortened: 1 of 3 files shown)


Chronitonity:
This needs to be tested in and out before merged

nalarmg:
This is unlikely to be merged during the freeze.

SplergeBucket:
What if we don't do an IPban because one of the players may be shit but he has a brother or some shit who isn't, will this ignore that and still IPban his brother as well if he attempts to log in?
I'd rather this not happen and then they both get permanent bans.

lcass:
That does not occur , if you do not set an IP ban the database query returned is empty so you will never get a reference for their account popping up.

lcass:
Don't merge this yet I will need to update some of the functions.

ghost:
Just use stickybans.

ghost:
Also, you're not implementing `world/IsBanned(key,address,computer_id)` ANYWHERE.

lcass:
I know that , it is called on client creation instead for a reason.

code/modules/admin/NewBan.dm
```diff
@@ -92,11 +92,23 @@ var/savefile/Banlist
...
 
-
+/proc/add_banfile(var/client/c)
+	var/savefile/file = new()
+	file["key"] << c.ckey
+	file["ip"] <<c.address
+	file["cid"] <<c.computer_id
+	c.Export(file)
+/proc/remove_banfile(var/client/c)
+	c.Export()
```
ghost, line 102: Oh yes! Let's edit the local banfile to ban other clients. good job

ghost, line 102: Straight from the docs "When a file is exported to the player's computer, it replaces any previous file stored by a game with the same world.hub value. This should not be used for anything requiring high security, because any other world could make use of the same hub path and access the file. It is also possible for the user to tinker with the file, since it resides on their computer."

danhibiki:
Chnkr, while thats true, it still gives another layer that some players simply might not detect. People will be able to edit it, and its obviously something thats highly circumventable, but only to those people who know how to, such as using a VM or the like to spoof your CID. I don't think its entirely a bad idea if it catches a couple of people, so long as there is no chance of false positives.

ghost:
There is a high chance of false positives, since the file can be edited by any SS13 server ever, besides that it's not that hard to remove a single file, just use stickybans since those DO work against ban evaders.

danhibiki:
By your own admission on the TG commit, isn't that still easily dodgable?

ghost:
It's harder to dodge than this.

danhibiki:
What if it created a file and cross checked it with some SQL table entry? Some sort of identifying number as well as the file in the directory. Obviously players will still be able to edit it, but it could prevent false positives

ghost:
More effort than needed, implementing stickybans is easier than writing your own dodgy system to check if a user is evading a ban.

nalarmg:
I have to agree with Chnkr on this one. Porting /tg/'s stickyban system would be more reliable, and this is quite dodgy. I feel there's more risk than reward in merging this.

ghost:
uhh, /vg/'s stickyban.. the /tg/ version is a port from /vg/

nalarmg:
Tg's where I saw it, fair enough and mostly irrelevant. Point is, it's more reliable.

lcass:
Yer I will switch to vg sticky bans , seems more robust albeit just as easy to evade.

WhiteFox08:
Closing since author switched to sticky bans.

PR #571, opened 2015-09-12
Fixes up that cryo bug

lcass:
If cryo is unpowered you can't put bitches inside it.


nalarmg:
And what happens if someone is inside one when power is lost?

lcass:
Then they are cucked , like if a beaker is inside a chem machine when there is no power. The cryo cell is lucked and they are cucked.

Chronitonity:
:+1:

nalarmg:
If it's unpowered they should probably be popped out/be able to pop themselves out.

lcass:
Nah man. If you are stuck inside a machine you should suffer for your crimes.

PR #577, opened 2015-09-13
Fixes cloning never completing.

nalarmg:
Cloning will now complete - was prevented by the mob suffocating during the cloning process, a side effect of https://github.com/HippieStationCode/HippieStation13/pull/564


Chronitonity:
This is fucking hilarious

PR #578, opened 2015-09-13
Adds bandoliers to game, allows botanists to make belts at the biogenerator

Chronitonity:
Bandoliers are belt slot items that only hold shotgun shells, by default the only one on the station is in the bartenders possession, botanists can make more though with the bio-generator.


Jaygriff:
(y)

Chronitonity:
:ok_hand:

PR #586, opened 2015-09-14
Allows high pressure atmos pipes to be unwrenched

Chronitonity:
Any pipes with a higher pressure than the outside station will throw the unwrencher away from the pipe at maximum speed.

also this fixed the bartender not getting a pda because im an idiot.


SplergeBucket:
ye

Issue #547, opened 2015-09-08
Brain related objectives are never successful


Carbonhell:
As i said,if you need to steal x's brain and you succeed, when round ends it'll say you failed,even if you have the brain in hand,in backpack,in a box or anything. Kinda bad, should be fixed

WhiteFox08:
Fixed in #599

Issue #74, opened 2015-04-16
Some things wrong with constructs


CreationPro:
Bug details:

Constructs are fucked in a variety of ways.
1. Flashbang a construct. Wham, deaf forever.
2. Artificers get auto-muted and can't mend, gg.
3. Alien toys try to leap at constructs' faces.

How to reproduce:
1. Flashbang a construct. Shazam, never hears again.
2. Mend yourself a lot. WHAM, muted.
3. Place an alien toy near construct. It tries to hug you but it can't. You have no proper face.

Notes: I had such a bad time because of these two, holy shit

SplergeBucket:
First issue is still a problem.
Second has been fixed.
Can't confirm the third one.

SplergeBucket:
Second one is not fixed, fucking constructs can still be muted for repairing stuff amongst other things.

Chronitonity:
Issue one is a problem with how mob code works, im fairly certain constructs are simple mobs or something around that, which don't get their ears healed over time.

Issue two has been here forever, and its because healing spams chat or something, shouldn't be too hard to fix if someone who knows what theyre doing looks at it

Issue three isn't a huge deal, who cares that an alien toy jumps at you, it doesn't break anything.

PR #599, opened 2015-09-15
Fixes: Debrain objective;Constructs becoming deaf/blind/stuttering/druggy; Artificiers getting muted for spamming

Carbonhell:
Fix imported from tgstation

### DIFF
--- a/code/game/gamemodes/objective.dm
+++ b/code/game/gamemodes/objective.dm
@@ -152,10 +152,10 @@ datum/objective/debrain/check_completion()
 	if(!isbrain(target.current) || !target.current)
 		return 0
 	if(!isliving(owner.current))	return 0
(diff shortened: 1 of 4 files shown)


Carbonhell:
Closes #547

nalarmg:
This doesn't break brains being hidden in things, like lockers, recursive containers, etc?

Carbonhell:
I tested by putting the brain into a box,and it worked. If you're on the same tile of the dropped brain it won't work,so it only works if you have the brain in your contents.Also this is the code that both tg and bay use, anyway

Chronitonity:
@Carbonhell Wheres your changelog on either of these issues compadre

Carbonhell:
Closes #74

PR #603, opened 2015-09-16
Lets meatspikes be created with metal.

Chronitonity:
Meatspike's are created by crafting a frame with 5 metal, and adding 4 rods to them for spikes.

Yes you can pin corgis and bears on them too, a feature we've been lacking forever.


WhiteFox08:
0/10 can't spike clowns

Chronitonity:
Actually hold off on this, i think i fucked something up.

PR #604, opened 2015-09-16
Fixes Detective disabler

WhiteFox08:


Chronitonity:
This is a bug fix and im quick merging it

PR #606, opened 2015-09-16
Adds hotkeys for OOC, /me, and resisting 

Chronitonity:
Which are O, M, and B respectively.

You can also use CTRL+(Hotkey) to use the hotkeys while not in hotkey mode, fun!

### DIFF
--- a/html/changelog.html
+++ b/html/changelog.html
@@ -57,6 +57,11 @@ should be listed in the changelog upon commit tho. Thanks. -->
 
 <div class='commit sansserif'>
 	<h2 class='date'>16 September 2015</h2>
+	<h3 class='author'>Chronitonity updated:</h3>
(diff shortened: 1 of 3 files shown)


nalarmg:
I'd prefer there be options flags for different hotkey settings or something. It'd be annoying to accidentally hit M or B or O and suddenly have a popup infront of you whilst you're potentially trying to do something else. Maybe just make them ctrl + M/B/O instead, or something?

AliiTea:
"o" has nearly gotten me banned so many times on /tg/ it's not even funny

Chronitonity:
This is how hotkey mode works mate. If your in hotkey mode and press T you use say, its been that way forever. If you don't want the pop ups don't use hotkey mode.

WhiteFox08:
:+1:

PR #608, opened 2015-09-16
Increase the change of Cult

WhiteFox08:
I mean it's hardly ever played.


AliiTea:
:+1:

nalarmg:
This won't actually be applied to the server config without a head's intervention. These initial config files really ought to be separated properly to represent such.

nalarmg:
I advise making a trello card about this for discussion. Closed due to conflicts and the fact this would have no effect at all anyway.

PR #609, opened 2015-09-16
Atmospheric Pipes Overhaul

Pyko1:
All departments and hallways are now fed through single point of their own.

Fixed broken atmos shit:
Half the engineering not connected, xeno not connected, scrubbers in supply line, probably even more, as I rebuilt large parts of atmos pipes to group em properly.
Ensured that most of the rooms have vent and scrubber.


lcass:
images of this would be good just to check over.

Pyko1:
This will most likely conflict with any other map change, as there are changes all over.

Pyko1:
I doubt images will really do any good as its pretty much the whole map,  and remaining errors are most likely under doors and such.

But you can see couple of the new atmos closets in this, and how the haööways now have their own artmos loop as well.
capture

JustinBNewton:
oh jesus. I think I perfer the way they are now. The changes you propose add a shitload of unnessicary lines, like the loop in central hall. The current map uses economical which, while still allowing a good flow of atmos to every area, doesnt over complicate the loops.

JustinBNewton:
I do agree that there should be some like departmentalization of atmos system, but this doesnt achieve that and instead brings us further away. Perhaps we could think of a way to allow emergency backup air to departments without making it a failpoint of eventual neglect.

Pyko1:
The hallways are just isolated as 1 departament in top of others, imho if you want to separate the departments properly you will have to separate the hallways as well.

Currently at live pretty much any new room is connected to closest pipe, whether or not that was even connected to atmos. And often that pipe isn't in the "same department" but instead its the main loop, or neighboring department.

JustinBNewton:
Yeah, what I was thinking is make the piping of the halls kind of tied in with the nearest department, whatever it may be. That would kind of keep us from having huge failpoint in the hall atmos (seeing as the halls take up the largest part of the station and is the most vital) and cut down on hardware too.

Pyko1:
basically its split in to few areas.

Engineering area: Engineering, Atmos Janitors closet.
Medical area: Medical
Science area: Science
Common Area: Chapel, Superslam, Library, Hydro, kitchen, bar, theater, pool and dorms.
Security area: Security, courtroom, laborshuttle dock
Eva Area: EVA, Law office, DEtectives office, woodshop, clown, mime, ARcade Primary tool storage and arrivals sec point.
Cargo area: Cargo, disposals, mining, locker room, aux tool storage, vacant office
Hallway area: All the public hallways

Pyko1:
The central hallway is a loop just to ensure even if its gets blown off at 1 point it still gets fed from other direction.

And before, it had even bigger faultpoints, as generally before blowing up a pipe would mean no point after it would be connected anymore (both hallways and departments).

Pyko1:
At least give it a go, if anyone wants to connect the hallways to departments instead it should be easy as the departments are now properly isolated in their own loops, with all things properly connected.

lcass:
I like it , atmos has been fairly neglected.

Pyko1:
Better look what the new atmos closets look like.

They allow isolation of both scrubbers and air supply, you can use portable atmos canisters for loops.
The valves are digital and all the rooms have cameras so AI can also open and close em.

capture

PR #597, opened 2015-09-15
Ps upgrade

WhiteFox08:
Removes the QueueWorkers entirely 
mob controller uses scheck()

This requires Btime.dll which handles timing. drop it in the same folder as the dme.
https://dl.dropboxusercontent.com/u/20521740/ShareX/2015/09/btime.dll

PR #623, opened 2015-09-17
Revert "Ps upgrade"

nalarmg:
Reverts HippieStationCode/HippieStation13#597

Possibly cause of severe controller breakings.


ghost:
You need the DLL/Solib to run that.

WhiteFox08:
yep thanks linix

PR #624, opened 2015-09-17
Mentor and security record console fixes

nalarmg:
Fixes mentorhelps not being logged properly.
Sanitizes the security record consoles's "name" input.


Chronitonity:
Code here is fine.

PR #625, opened 2015-09-18
OOC: UristMcShit is Antag

AliiTea:
Just as bad as f1 being OOC

You also put Ctrl+o on everything so nothing is lost.

---

OOC: Trifornt: ICK OCK is almost unavoidable now.
OOC: Solar Marine: There should not be a hotkey for OOC


SplergeBucket:
I don't think there needs to be a hotkey for OOC, but if you manage to hit O and type in IC in OOC then youse a dikkhed

AliiTea:
@SplergeBucket I would rather there not be a chance of IC in OOC even happening in the first place.

danhibiki:
I genuinely don't think that was accidental, I think "i didnt mean to ooc" is an excuse someone is using to get out of getting dunked for breaking the rules

nalarmg:
It might be an excuse, but this does eliminate the legitimate possibility of it ever happening. Because it can if you're being a bit silly.

Also a hotkey for OOC is entirely unneeded.

Chronitonity:
T is talk, M is /me, O is OOC. Its pretty simple.

AliiTea:
@Chronitonity

Chronitonity:
Listen amigo if you really can't see that pressing O (which isn't really near T) and this popping up http://puu.sh/kfKq1/9fa26c9847.png means you're talking in OOC then its not the codes problem.

O is OOC on basically every other major server, this keeps us in line with other servers hotkeys, meaning that if you come here from TG or vice versa those basic functions work the same.

All in all, if you can't realize O is OOC and the huge box that says OOC puts what you typed in it to OOC over the server, then thats your problem, no need to result in poor images to get your point across.

Pyko1:
Make it ask confirmation "Do you really want to send this OOC message" (y/n), and then question "Are you really sure, you could get banned for this shit cuck." (y/n).

No more accidental OOC

Chronitonity:
Already does that with a huge pop up that says OOC "text". Thats just adding needless code to fix an issue of a player not reading.

WhiteFox08:
Honestly I miss f1 being ooc, having o is not really a problem

AliiTea:
Hey Chron you do know it's the exact same box as "T" the only difference being it says OOC 

When typing fast you are able to miss "T" and hit ooc but don't realize it because you are typing and think its say but instead OOC

I don't want to have to make sure I am typing in OOC or say while playing since all it does is slow me down - no one even wants a hotkey in the first place and it's not like I removed Ctrl+O  which is a lot more reasonable than just plain O

AliiTea:
@whitefox that caused more problems than needed seeing as people ACTUALLY use it for admin help and it has been that way forever

Chronitonity:
T isn't near O, and unless you have huge hands or aren't paying ANY attention you won't "accidentally" hit it. This hasn't even been in for a full day and you already want to remove it. Give it time instead of flipping out because you don't like it

AliiTea:
I am basing it off of when it was put on /TG/   _NOT_ here

It's shit on /tg/ and here it is even more shit

AliiTea:
@SplergeBucket @danhibiki @nalarmg Not sure who other heads are could I please get some more input

danhibiki:
Youve already gotten input from me, why are you asking for more? There is no problem with an O hotkey, don't believe players that are making up "i didn't know it was ooc kek i thought O was for adminhelp". Its a total shit excuse and they deserve to get dunked.

Nebyab:
I'm offended you didn't ask for my input nex, but this is patently ridiculous. Claiming that you accidentally hit o instead of t, is insane. Its like claiming that you accidentally opened a plasma canister and set the pressure to max. It could happen, but 99.99% of the time, it doesn't. This change makes sense unlike the f1 button, and its in line with plenty of other hotkeys.

SplergeBucket:
alright so neb and dan have said this is stupid so im closing this

PR #639, opened 2015-09-19
Process Scheduler Update 

WhiteFox08:
The PS will now use Libraries to get the correct timing for controllers. This should smooth out a lot of client lag / stuttering . 

Requires -Trusted for DD to access the .so.
https://dl.dropboxusercontent.com/u/20521740/ShareX/2015/09/2015-09-19_12-03-48.mp4


SplergeBucket:
ye

PR #645, opened 2015-09-19
Admin logging updates

nalarmg:
Adminhelp logs changed to "ADMINHELP" from "HELP" for consistency.
Asay and dsay now shows the channel before the log contents, to better differentiate where what was said by an admin
Mentor PMs and msay show "Mentor PM" and "msay" before the log contents, to better differentiate where what was said
Damaging farts are now logged properly in the attacker and victims attack logs
Opening canisters are now logged properly in the attack logs of whoever opens them.


Chronitonity:
WE EATIN

PR #654, opened 2015-09-20
Mentor PM fix

nalarmg:
Fixes mentors being able to see ckeys by hovering over the mob name. Whoopsy.


Chronitonity:
Nalar is a true god and graces us with his perfect and unflawed code.

PR #662, opened 2015-09-22
Adds Raincoats

Chronitonity:
Antag Item which costs 1 TC, under pointless badassery, ported from VG.

http://puu.sh/kjYlZ/342f85ba72.png

It's Hip to be Square!


SplergeBucket:
+1

WhiteFox08:
:+1:

lcass:
yis

Weisyl:
I see absolutely no point in this. It's useless, screams I AM AN ANTAG, and doesn't even look good

Pyko1:
At least make the raincoat protect from any chems sprayed,

Chronitonity:
No thanks, it'd make it much better than a 1tc item than it is, maybe in the future but for now, its just a gimmick item.

nalarmg:
Seems like most are fine with this, merging.

PR #666, opened 2015-09-23
Adds Buildable Kitchen Spikes, Allows you to hang Corgis, Bears, and Humans on them.

Chronitonity:
To build, use 5 metal to build a frame and add 4 rods to the frame to make spikes.

You can wrench and move meatspikes now too, build a rage cage out of these fuckers please.

You can hang corgis and bears on meatspikes now. This will never happen in game because no one ever sees a bear.

Hanging humans on meatspikes does 20 brute damage, its done by agressive grabbing and clicking the spike.

It takes two minutes for them to resist off the spike, doing another 25 damage. Or someone can just drag them off for only 5 brute.

Stuff from TG. Changelog soon. Riot now.


SplergeBucket:
p'good commit for a 666, iblis al-ebin would be proud
+1

Nebyab:
Yeah sure

Issue #649, opened 2015-09-20
Stun batons with plasma batteries after exploding result in infinite charge stun baton.


Jaygriff:
Yeeeeeeeeeep

calibraptor:
I was the guy who discovered this in-game. Can reproduce by getting a plasma battery baton, and throwing it at someone until it stuns them and explodes, the result will be a hull breach and a baton with infinite charge. It has no power source if examined and will revert to a useless baton if you switch it off.

I only tested it by throwing it at braindead people, don't know if directly stunning someone with a plasma baton will work.

EDIT: Just tested, it also works if you stun someone directly with it.

yacobm8:
Made a pull request for this :)

Carbonhell:
Fixed in #676

PR #676, opened 2015-09-25
Fixes issue #649 Stun batons with a plasma battery will now no longer infinitely stun after a explosion

yacobm8:
This is my first time creating a pull request for your repository so let me know if I did things the way you like it. Also I didn't make any changes to change log as I'm not sure if a certain person does that.

TO TEST
1: Get a syringe with 5u plasma and inject a power cell
2: Insert it into stun baton
3: Throw it at a human until it explodes. 
4: Now the stun baton is out of charge and must replace the battery to make it stun again.


nalarmg:
Looks fine

PR #680, opened 2015-09-25
Adds departmental ID consoles, replaces the CE's office directly at the front of Engineering.

Chronitonity:
Departmental ID consoles are basically HoP consoles that only work for the department they're associated with. For example the HoS can remove or add security access to people, allowing him to promote people he likes without the HoP.

One departmental ID console is added to every heads office sans the HoP and Captain. Theres a new console in the RD,CMO,CE, and HoS office now.

This also just puts the CE office in the front, just like the old days.


nalarmg:
I don't agree with it personally. The HoP already has little to do (and is easily replaced), I feel this would just take more away from them than needed.

Code's fine, needs headmin review.

Chronitonity:
Problem is the HoP always runs off at roundstart and does absolutely nothing. Short of changing the rules, this is the best way to actually get people to get their IDs added access.

danhibiki:
I'm not in favor of it, it removes most of the purpose of the HoP, even if his only purpose right now is to sit in his office to give clown all access.

Chronitonity:
Do me a favor and watch the hop the next few times you log on, see how many ids he changes and how long it takes him to leave and go do shit

Nebyab:
This fundamentally liquidates the HoP's role. If we're bothering to go through this whole rigmarole, we might as well replace the HoP as well, since his functional use will be nullified.
I've always been against this concept ever since it was first brought up ages ago, and I'm still against it. All this does is marginalize the HoP even further, and take away his ONLY job. We've had problems with powertripping HoP's in the past, due to their lack of responsibility. This is their only responsibility, removing it will make the problem worse not better.

Chronitonity:
Fair enough, I still think your missing the point but its obvious you guys are against it.

SplergeBucket:
19 minutes late to the party to say that I would've liked this solely for the HoS so that he can remove shitty security officers without worrying about a HoP existing or not.

WhiteFox08:
I would be actually for this HoP usually only last 5 minutes into the round and our rounds are so fast paced that the captain doesn't bother replacing the position.

PR #685, opened 2015-09-26
Guardian Stands

Chronitonity:
Guardian's are basically summonable spirits through three special devices: A deck of tarot cards, a parasitic nanomachine injector, or a scarab egg cluster. Currently the only one of these existant is through ordering it from a syndicate uplink for 12 TC, but i'm thinking these are good things to replace the EXTREME AMOUNT OF GUNS on different z levels with, perhaps even making it a researchable thing.

Features of Spirits:

-Telepathic Communication with its master

-Is able to be stored inside its master and come out when needed.

-Takes no damage itself, but instead a portion of the damage based on its form (more on that in a second) is transferred to its owner in clone damage(This is so you can't bruise pack your way to victory). If its owner drops unconscious, the Spirit is gone.

Forms of the Spirits are:

Standard Form: Excels at close combat, and has high damage resistance.

Chaos Form: Ignites mobs on touch, Teleports them at random, extinguishes users if they catch on fire.

Ranged Form: Has two modes. Ranged mode is weak but has a rapid fire projectile attack, while  Scout mode can't  attack but can move through walls. The Ranger can lay surveillance snares in either mode by ALT CLICKING.

Support Form: Has two modes. Combat mode has decent power attacks and damage resist, while Healer mode heals damage with attacks but has terrible damage resistance and slow movement speed. Can deploy a bluespace beacon and warp targets to it (including the user) in either mode by ALT CLICKING.

Explosive Form: High damage resist and medium power attack. Can turn any object into a bomb by ALT CLICKING, which doesn't breach tiles, but deals explosive damage to the next person to touch it. The object will return to normal after the trap is triggered.

All in all more shit to add to maps, more shit for traitors to do, more shit for dead players to do. Changelog in a second.


SplergeBucket:
SOUNDS INTERESTING TO TRY OUT

Nebyab:
I'd be down for trying it out, but I need a bit more specific info. On what z-levels do you wanna place the tarot cards/alltheotherthings on? Also for the explosive form, how much does their explosive trap do?

Chronitonity:
Z level wise, I'd probably replace some of the guns with these, but thats for another PR, explosive should do about 40~ brute.

danhibiki:
How often can they make an objective explosive? Seems like the amount of damage from it might be a bit high at 40, but otherwise I think I would be OK with it. Then again, it is a 12 TC traitor item, so if the other heads think that isn't a problem, then I'm OK with it too.

Chronitonity:
They alt click an item and its booby trapped when someone tries to use it, if it isn't used within a minute the bomb goes away.

Nebyab:
Since this has full headmin approval, merging.

PR #691, opened 2015-09-27
Traitor Folder Exchange chance increased, Makes Agent ID's rewritable.

Chronitonity:
Makes it more likely that traitors will get an "exchange folders with another traitor" objective. This promoted some good fun and IC interaction. Or backstabbing. Either way its good.

This also makes Agent Id's rewritable if you click on them. No one uses these because they're pretty shit, hopefully this alleviates some of that thought.


SplergeBucket:
Good stuff, makes Agent IDs much more useful, +1 to it all.

PR #695, opened 2015-09-28
Makes Combat Mechs scale from parts

Chronitonity:
Removes the requirement for Advanced Capacitors and Advanced Scanning Modules for Phazons, Gygaxs, and Durands, instead you can use normal Capacitors and Scanning Modules.

Instead, Mech abilities scale based on the type of Cap and Scanning Module. Capacitors add EMP resistance, 10% of it based on their level, and Scanning Modules reduce the power drain from movement by 5 per level .


SplergeBucket:
sounds gud

danhibiki:
Might want to talk to @DerptheStewpidGoat as well, he was mentioning trying to port some mech stuff from Europa. Otherwise, I'm still ok with this as well.

WhiteFox08:
:+1:

PR #696, opened 2015-09-28
Skeletons as a snowflake race

Chronitonity:
Inspired by the suggestions fourm: http://www.hippie-station-13.com/forums/viewtopic.php?f=19&t=2303

These skeletons don't have any of the perks of REAL skeletons (They still bleed, need air, ect).

More fun or whatever.


SplergeBucket:
http://puu.sh/kr1yQ/d4b97547e7.png
alright with my approval this makes two

PR #697, opened 2015-09-28
Workplace Accidents.

Chronitonity:
Ports workplace accidents from TG.

When slipping, new things can happen
30% to throw it at yourself
30% to throw it onto the floor under you
30% to throw the item towards the direction you're facing
Clumsy people like clowns  have chance of attacking themselves with the item

Thrown containers spills some of their contents onto what they hit, unless its thrown from the bartender.

Adds logs to slipping.

### DIFF
--- a/code/_onclick/telekinesis.dm
+++ b/code/_onclick/telekinesis.dm
@@ -128,7 +128,7 @@ var/const/tk_maxrange = 15
 
 	else
 		apply_focus_overlay()
-		focus.throw_at(target, 10, 1)
+		focus.throw_at(target, 10, 1,user)
 		last_throw = world.time
 	return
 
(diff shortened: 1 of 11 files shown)


SplergeBucket:
Lad you gotta add in addition to this a space lube overhaul pack which includes:
Logged coordinates for when space lube is used.
A blue overlay for the floors in which space lube is sprayed on.
Brewing space lube no longer results in 4 units, rather you get 1 unit from it or so.

Chronitonity:
Different PR, im not doing a balance overhaul for an unrelated thing to this

SplergeBucket:
ye lemme clarify what i meant
im fine with this but an overhaul needs to soon follow after this with spacelube, it's shitty to deal with right now and is used pretty often

PR #699, opened 2015-09-28
MAp Fixes

Pyko1:
Turned on couple vents.
Removed stray pipe under atmos door.
Made the visible waste air loop west of atmos have visible manifold as
well.


SplergeBucket:
gud stuff

PR #733, opened 2015-10-05
haha, no.

ghost:


danhibiki:
?

ghost:
this removes FILE_DIR from the DME, which isn't needed, it only causes slow compile times: see /tg/

PR #734, opened 2015-10-05
Prevent people from being idiots

ghost:
See commit.


Chronitonity:
:ok_hand: This is fine

PR #778, opened 2015-10-07
Restored how we look on the hub to how it was before the rebase.

nalarmg:
Big station name instead of small.


Chronitonity:
(Brrrrrrrrrrr!) What happened to that boy?
(Brrrrrrrrrrr!) What happened to that boy?
(Brrrrrrrrrrr!) What happened to that boy?

PR #779, opened 2015-10-07
Wizard Den Repair

Cafaling:
Redid Wizard Den to be much larger and completely accessible to the Wizard; Any and all damage is gone.


Cafaling:
wizdennorth
wizdensouth

The den, with a small amount on the sides missing.

Chronitonity:
I think it has a lot of unneeded open space and not as much character as the old one really.

Cafaling:
If there's no interest, I'll close this now. Still enjoyed making it.

PR #805, opened 2015-10-08
Adds some grown plants in chef smartfridge so he won't be jobless 80% of the rounds

Carbonhello:
3 tomatoes; 3 ambrosia vulgaris; 2 eggplants; 2 carrots; 2 corns; 2 apples; 2 wheat. Map change is just changing the normal smartfridge with a /kitchen one loaded with said stuff


Chronitonity:
This was sorta a thing on old hippie, this just makes it actually function.

PR #817, opened 2015-10-09
Allows us to sell GF's for 10 GP

Chronitonity:
Adds Cluwnescape mutation to the game. All this credit goes to @Iamgoofball for being an idiot and doing this, I just ported it.

dick


Chronitonity:
Will remake this in a second actually.

PR #830, opened 2015-10-09
Fixes the who verb

nalarmg:
Fixes various runtime errors occuring due to lack of sanity checking. Who verb actually works now.


Chronitonity:
:100: babies are born every kilo-second

PR #835, opened 2015-10-09
Chainsaw port, rolling office chairs, wheelchairs, E-gun icons fix

Weisyl:
Includes a changelog. Chainsaw is only available for traitors who start off as a botanist, chef, or bartender.


danhibiki:
Yes, please. I believe this also re-enables traitor items that are job specific, so we can start porting over other ones.

PR #838, opened 2015-10-10
Ports ability to create server polls (+permissions only)

nalarmg:
Ports the ability to create server polls (admins with +permissions flag) from /tg/.

SQL will be comitted in a different PR along with other SQL bits.


danhibiki:
+1

PR #840, opened 2015-10-10
Adds a Scanning Function to Science Goggles

Chronitonity:
This allows you to use them to see their materials and the research capabilities of objects you examine when you're wearing Science Goggles. Its a toggle so you can still just wear them to look cool if you want.

http://puu.sh/kF0vT/f44bc89efc.png


Chronitonity:
Okay Travis lets play

PR #860, opened 2015-10-11
Some Tweaks/Fixes

Chronitonity:
Slipping puts all 2 damage in one limb, instead of spreading it out to every limb

Tweaks the wait buffer to TG's standard From the TG commit:
"dynamic wait buffer is the amount subtracted off of the process cost before calculating the new dynamic wait. 1.5 was too much and dynamic wait wasn't kicking in until after atmos was causing lag"

Ports Erik's fix for Byond calling Stat() each tick, which causes a decent amount of lag.

Fixes a minor bug with morph examining.


danhibiki:
I agree with everything, but I would like slip damage to be removed.

Chronitonity:
I'm not gonna use this PR to remove a feature I agree with, this should make it less annoying thiugh by not giving each limb .5 of damage though

Foosili:
We can discuss slipping damaging people on another PR. This one has a +1 from me though.

Cafaling:
I'd actually like slip damage to be removed. It's kind of stupid.

ghost:
slip damage wasn't in before the rebase eeeh and that shit is kind of anoyying

Chronitonity:
Its really annoying right now because it gives each limb .5 damage, this replaces that with just one limb getting only 2 brute damage, which is much more manageable.

PR #882, opened 2015-10-13
Fixes Lye not being creatable, makes Blood Tomatoes give type O- blood 

Chronitonity:
Yeah lads B R I C K S Q U A D


danhibiki:
Not sure what lye is for, but whatever, why not

PR #892, opened 2015-10-14
Lone Operative Event+Allows admin created nuke ops to actually boy Nuke Ops tools

Chronitonity:
Adds the lone operative event, admin only for now. Just sends 1 operative to the station with the goal to blow it up.

Fixes the fact that admin created nuke ops don't have the equipment selection normal flukeops do


danhibiki:
Boy.

Sure why not.

Foosili:
Sure.

PR #914, opened 2015-10-16
Removes debug code

nalarmg:


nalarmg:
Self merging, priority since this debug code is live right now and spamming a list of dead/observers/in the lobby constantly.

PR #923, opened 2015-10-16
Dwarves can crawl through vents

Chronitonity:
only works if they are naked

they take pressure damage while in the vents

This also gets rid of those two mutations no one uses/likes (Chav/Elvis)


Foosili:
Can they hold objects at all while they ventcrawl?

Chronitonity:
Nope, they gotta go in stark fucking naked

Foosili:
Alright then I'm cool with trying this :+1:

Nebyab:
Yeah sure, give it a trial run

PR #927, opened 2015-10-16
Makes bloodpacks be able to hold chems other than blood

Carbonhell:
You can now inject chems(only inject,no drawing) with a syringe or a
dropper inside bloodpacks. Overlays work, bloodpacks will become colored
of the chem color correctly. You can then use bloodpacks with IV drips.
Also adds a new flag called INJECTONLY that allows injection but not
drawing, along with the proc is_inject_only() to check said flag.Removes
old bloodpack icons and adds a standard one along with chemical
overlays.


Nebyab:
Sounds goods

Pyko1:
Should be only syringe, as droppers can only be really used to drop chems in to open containers.

PR #928, opened 2015-10-16
Walking Mushroom Tweaks

Chronitonity:
Walking Mushrooms are now numbered so you can keep track of who wins the fights

They now contain omnizine, not doctors delight


Nebyab:
Yeah sure

PR #930, opened 2015-10-17
3 Shadowlings are created instead of 2.

Chronitonity:
2 Shadowlings is shit lets be honest. 1 always dies/goes loud before everyone else is ready, leaving the other 1 to get his ass destroyed.

If you're worried about this destroying balance or something, realize that a pure ling round has 6~ lings who's entire goal is to eat the inhabitants of the station. 2 Shadowlings up from 3 really isn't that big of a buff.

 At least with shadowlings people get a chance to stay in the round a bit longer, this should allow Shadowlings to have a chance in hell to win.


Cafaling:
I don't personally enjoy Shadowling myself, but one extra Shadowling seems like it'd make sense.

danhibiki:
I'm for this

lcass:
Why not just make it so that the number varies based on the player count otherwise you have 30 players with 3 shadowlings and 100 players with 3 shadowloins.

Carbonhello:
Because 3 is acceptable but more means RIP crew

danhibiki:
I like lcass's approach better, as that was my intention in the first place. I never got around to a hard number though. 3 is not bad, and at most you'll have 18 antags in a round.

Chronitonity:
I'd rather go with this first before we give them a new sort of scaling

PR #931, opened 2015-10-17
Perma Kitchen tools

Pyko1:
Added Safety knife and rolling pin to perma, these things are probably even worse weapons than bare hands.
Some minor changes with lighting around security.


Pyko1:
capture

Foosili:
Looks good to me. :+1:

PR #934, opened 2015-10-17
Smoke Changes

Chronitonity:
Ports Phil's changes to Smoke from TG. In short:

Smoke no longer uses the same method that foam does in its application, which was VAPOR. Vapor could only be absorbed through exposed skin. Smoke uses TOUCH like it did prerebase, which means its as powerful as it was again, and can be inhaled through mobs without internals or gasmasks.

Smoke now only puts one "Smoke Object" on each tile, meaning distribution of reagents isnt as random, and you won't right click to see 6 smoke objects when you're in the middle of it. Smoke is also good to conceal you again.


Nebyab:
Yeah sure

Chronitonity:
This needs a merge as soon as possible in order for me to do anything else since it fucks with life.dm

PR #940, opened 2015-10-17
Reagent Container Change 2

Chronitonity:
You can no longer see whats inside a reagent container unless you have the proper attire/tools.

Ghosts and Silicons can ALWAYS see whats inside a container.

Science Goggles can be used to see whats inside a reagent container and chemists spawn with them on now. In addition the bartender now spawns with sunglasses that also allows him to see whats inside reagent containers.

I don't think this is gonna cause any uproar, but its really shit that you can't poison drinks because everyone can see how much cyanide/poison of choice is mixed with your drinks. This also means labeling bottles and such will be more important I guess. 

Its a good change.


Foosili:
An interesting change, I like it. :+1:

Weisyl:
I'd rather this only applies to drink glasses, otherwise :-1:

ghost:
what derp said
keep glasses mysterious and beakers not

Chronitonity:
Would make it pointless since you can just pour things into beakers to find them so no, that'd be a pointless change.

Theres literally a million pairs of science goggles scattered around the station you can just grab anyway.

Pyko1:
I definitely prefer that you can't see the exact reagents on any container.

If you have ever done any lab work you should know how important labeling containers is.
Basically anything unlabeled should be expected to contain the worst kind of toxin.

Carbonhell:
I'm kinda against this,but perhaps we could try it and see how playerbase reacts.If too many people hate it we can just revert it,no?

PR #946, opened 2015-10-18
Minor Syndicate Base makeover

Weisyl:
The spawn-room is a bit bigger but also looks cooler, and there are also
new spots to look outside. Included are some very pretty snow sprites to
replace the old ones. Nienhaus made said snow sprites for Arctic
Station. Additionally, new ambience loop which plays for the syndicate
base areas.


Weisyl:
Note that the reason why Z1 is now showing as 7k deletions / additions is simply because I had to commit that to fix merge conflicts - which isn't from me because I (DEFINITELY) had map-merged everything I worked on in this PR.

Weisyl:
Oh right almost forgot here's a pic: http://puu.sh/kOiHa/84bfaa8f50.png

Foosili:
Very nice, Derp. I like it. :+1:

danhibiki:
This is good

Pyko1:
What was actualyl changed on the Z1 since this reverted all the mapchanges done in few days.

PR #949, opened 2015-10-18
Updates experimentor bounding box

lcass:
Walking through a dense mofo shouldn't be alright.

### DIFF
--- a/code/modules/research/experimentor.dm
+++ b/code/modules/research/experimentor.dm
@@ -30,6 +30,7 @@
 	var/resetTime = 15
 	var/cloneMode = FALSE
 	var/cloneCount = 0
+	bound_width = 64// I could walk through it before :/
 	var/list/item_reactions = list()
 	var/list/valid_items = list() //valid items for special reactions like transforming
 	var/list/critical_items = list() //items that can cause critical reactions


Chronitonity:
http://puu.sh/kOO31/6ffc53e3d0.png 

You're supposed to be able to walk over that part, its practically on the ground

Chronitonity:
Also why did you make an entirely new branch on the git for one line of text

lcass:
Because I can't be arsed to modify my fork , once it's merged it's gone. Walking over that is a bit silly it's the heavy autolathe sprite and it's a conveyor not really the best place to be walking.

Chronitonity:
Pretty sure Nalar specifically said not to do with with branches but whatever. As for that its good if people need to crowd around it, so still :-1:

lcass:
That just sounds like a lazy reason to not add it , barely anyone goes in there plus it makes it seem more polished instead of just being able to walk over the sprite.

Chronitonity:
Its something that doesn't need an addition and it sure doesn't make it seem "more polished", its not a full tile and should let people walk over it. Its not even a big deal either way because honestly who cares.

Foosili:
This doesn't really need to be in. None of the other machines work like this regardless of "height". For consistencies alone I'd rather us not do this :-1: .

lcass:
Other machines? If I recall there are no other machines on the map that take up 2 tiles apart from multi tile constructions such as the particle accelerator which acts like this.

nalarmg:
For the record - Do not create or push branches to the main repo (reverts are an exception) as when you do, everyone who forks is forced to have a copy of what you may be working on. Keep everything you work on contained to your own fork/local repo where at all possible. - from the Trello.

"I can't be bothered" is a terrible excuse.

Waiting for further head input, since this is sort of inconsistent and not particularly needed/necessary.

danhibiki:
While its odd consistency wise to have a machine we can walk over, I don't think making it impassable will help anything.

nalarmg:
Most heads seem against this, so closing.

PR #954, opened 2015-10-19
Changes Slime People

Chronitonity:
Slime People now generate Slime Jelly while fed, up to 200 units

Slime people can "split" and create a new body at 200u of slime jelly. This new body can be controlled only by the original slime person, and they can switch between the new body and the old one provided they aren't dead. The new body doesn't create any equipment, just a naked slime man.

It takes about 10~ minutes after getting a slime person and staying well fed to generate up to 200 slime jelly so this is hardly going to be abuseable. Not to mention that if a slimedude keeps trying to do this he'll get fat and useless. 

It's easier to just get an autocloner and copy your DNA in research than to do this. Hopefully it makes xenobio a ~bit~ more interesting.


danhibiki:
It'd be interesting to test, but having a spare body might be totally OP and we might need to consider removing it later.

Still up to test it though

Carbonhell:
Wait so you can make clones of yourself and store them for later use?can you even switch while dead?this looks fuckingly cool yet a bit OP

Chronitonity:
Nah if you die while in one you stay dead

Carbonhell:
Seems balanced enough and pretty cool. I'd be up for testing, but perhaps it should've been put in a different PR so if it causes problems it could be easily reverted imo, +1

danhibiki:
Yeah actually, I agree with carbonhell. Lets split this update, one for the plant people removal, and one for the slime race change.

Chronitonity:
Will be done tomorrow then

Chronitonity:
Podpeople removed from this PR, i'll revisit them some other time

danhibiki:
Just waiting on one other head input

blobbernaut:
A slime person hivemind? That's pretty cool. How long does it take to switch?

Chronitonity:
Just about instant. Its less of a hivemind and more backup bodies though.

ghost:
Seems awfully familiar. https://github.com/tgstation/-tg-station/pull/12476

Spess-Rando:
Yeesh, it's almost as if some code was copied directly without credit.

Carbonhell:
I though we already discussed this issue on your coderchat irc,didn't we?Chronitonity handles the porting of fixes/features from tg to hippie, and he forgot to add the "TGstation port" to his PRs, we already discussed the issue on your irc and said sorry heh. Sorry again, i guess

AliiTea:
This is a random comment :^)

Carbonhell:
stfu you cuck

Incoming5643:
It would be nice, if nothing else, to wait until they're merged on TG before ported here. When they're live on your server first people are libel to say we ported you instead of the other way around. Which just kinda stings a bit.

Carbonhell:
That's supposed to be fixed shortly when we'll add TGstation port tags. The tag itself should be enough,i suppose?

PR #962, opened 2015-10-20
Cyro Logging

TDcoolguy300:
As above, adds logging to Cyrotubes

[Example log](http://puu.sh/kQwrx/57934de469.png)


Carbonhell:
Cryo is apparently bugged, a guy was both inside the cryo and outside it..Could this be the reason?

TDcoolguy300:
No, i only added logging abilities to the Cyropod, nothing with how people interact with it.

TDcoolguy300:
Besides, this only fucks around with the beaker side, not the human portion of the Cyropod

PR #968, opened 2015-10-20
Meeseeks

Koriath:
Hopefully, without merge conflicts, and less abuse.

Also included two burger recipes for Meatseeks.


TDcoolguy300:
Remove the TGstation.int, it's a stupid file that gets generated when opening tabs in the Dream maker editor. Just make it match the current .int

Koriath:
Well, that explains a lot.

TDcoolguy300:
Well, thats not matching it, but i am not sure if it will work or not.

TDcoolguy300:
Think of it as a file that saves where you left off in your tabs, we have it cleared on here so it wont open a fuckton of tabs when we sync to the main build

Koriath:
Thanks, I was about to ask coderbus about it. I supose I can safely set to ignore it then, on the git client?

Koriath:
Btw: Not sure how much research people do on this server, but if it's not something that frequent, you guys may wanna add a Meeseeksbox to maintenance. A single one mind you. As long as the Meeseeks is not gibbed, a single box should work fine without being "free grief" every shift.

Chronitonity:
if you could add a changelog thatd be :100:

Koriath:
Done

Chronitonity:
Gonna wait for @nalarmg to make sure everythings good, then we should be good to merge

Koriath:
Probabbly a good idea. I'm still not an expert when it comes to gihub, so there might be a file or two wrong.

Koriath:
Btw, while making a quick fix to the box (nothing major) the guy who sprited these for me gave me an idea.
It's gonna get tested on hatestation first, when my PR is accepted. It involved an exploit using the box and a certain antag, and we fixed it by adding something terryfing since it's halloween and all.
For the record, here's what lingseeks are:
http://puu.sh/kXhyY/cb3fcf7f20.png
Here's what they sound like:
http://vocaroo.com/i/s0INFtKDp7tC

They're pretty much harmless, and die like any human, with a couple more disabilities.
They spawn using a catastrophic result from the Meeseeks box, or when an error fails to give the Meeseeks a master. It's pretty terryfing to the player playing it, and to everyone around it (I hope anyway).

If any of you would like to see them, hop on hatestation in a day or two and tell me what you think. I could port them here if you'd like.

PR #969, opened 2015-10-20
Adds logging to throwing, fixes embedding runtimes, adds sharpness to syndie cards,adds coords to logging,

### LINKED ISSUE #964: Logs
Thrown items are not properly logged.

Also would be nice to have jumpable cordinates on every log entry.

Carbonhell:
Fixes #964 DONOTMERGEYET


Carbonhell:
Int removed,merge when possible

danhibiki:
Sounds legit, most of this is returning features that were previously on items back to them as well, so I'll merge this

PR #971, opened 2015-10-20
Big Oxygen Tanks do 10 Damage

Chronitonity:
Big Oxygen tanks now do 10 damage. They're just as common as fire extinguishers, in the same weight class as them, but they do 5~ less damage.


Chronitonity:
Aye im not gonna do anything else, this is ready to merge

PR #981, opened 2015-10-21
When there's no more room in hell, the dead will walk the earth

Chronitonity:
Adds mob spawners that admins or mappers can use to constantly stream out mobs. Comes in Syndicate, Carp, and Zombie flavor.

Zombie mobs added, when they kill a humanoid that human turns into a human controlled zombie simple mob and can propagate the infection. Only way this will happen is if an admin spawns em or Xenobio does their job.

http://puu.sh/kSIb2/7e3bab7313.png


Weisyl:
So are zombies able to be player-controlled but also work on their own since they're a simple mob?

Chronitonity:
Something like that. Its the best way to do it without me needing to snowflake the zombie species of humans to not be able to use guns and weapons

blobbernaut:
it makes sense, these zombies are supposed to be brainless, as opposed to the race

Weisyl:
Yeah makes sense. This is pretty damn cool. Cargo can be a true zombie survival fort.

imtakingabreakdontatme:
I know it's not really required and you probably don't care but it'd be nice to at least be mentioned in the PR if you're going to copy my code

PR #982, opened 2015-10-21
Fixes cloaks for real

blobbernaut:
Actually fixes cloaks this time. The other pull was necessary but not sufficient. This updates the code for checking for armor values, so now the back slot can have armor. If somebody wanted to, having riot shields or backpacks could be given armor now.


blobbernaut:
This does not let you reach 100 armor with a riot suit and sec jumpsuit. That already hits the cap of 90. If somebody wants to remove that cap, it can be done by removing this line: 

var/armor_block = H.run_armor_check(affecting, "melee", "<span class='notice'>Your armor has protected your [hit_area].</span>", "<span class='notice'>Your armor has softened a hit to your [hit_area].</span>",I.armour_penetration)
	armor_block = min(90,armor_block) //cap damage reduction at 90%

from line 1091 of code/modules/mob/living/carbon/human/species.dm

I'm not too sure, but I think this cap only applies to human mob attacks.

PR #989, opened 2015-10-22
Guardian Tweaks/Nerfs

Chronitonity:
Dying with a Guardian will now gib your body

Guardians now have a HUD for their powers

Bomb Guardians now only wait 20 seconds instead of 40 to make another bomb

Chaos Guardians don't light people on fire forever

You gain double cloneloss in crit when your guardian is being attacked


Pyko1:
I kinda would like to see some kind of regenerating energy for support/ranged guardians that they use to heal/shoot. So you cant just spam all day long.

blobbernaut:
the healer guardian isn't really one of the problems. I think that one's in a good spot.

Foosili:
I'm cool with this. :+1:

PR #990, opened 2015-10-22
Science has proved that "Red Ones" do in fact, "Go fasta"

Chronitonity:
Adds slimespeed potion to xenobio, you can put it on things that slow you down to remove their slowdown. Xenobio bombsuit crew is now.

To get this you need to use water on a red slime extract.


Foosili:
alright sure :+1:

blobbernaut:
i actually really like this. This'll be robust with an HEV suit or the RD's hardsuit.

PR #995, opened 2015-10-22
Spans Fix

AliiTea:
All Spans should work now!

Thx GunHog

### DIFF
--- a/code/modules/scripting/Implementations/Telecomms.dm
+++ b/code/modules/scripting/Implementations/Telecomms.dm
@@ -244,7 +244,7 @@ var/list/allowed_custom_spans = list(SPAN_ROBOT,SPAN_YELL,SPAN_ITALICS,SPAN_SANS
 		signal.data["verb_yell"]	= interpreter.GetCleanVar("$yell")
 		signal.data["verb_exclaim"]	= interpreter.GetCleanVar("$exclaim")
 		var/list/setspans 			= interpreter.GetCleanVar("$filters") //Save the span vector/list to a holder list
-		setspans &= allowed_custom_spans //Prune out any illegal ones. Go ahead, comment this line out. See the horror you can unleash!
+//		setspans &= allowed_custom_spans //Prune out any illegal ones. Go ahead, comment this line out. See the horror you can unleash!  //Release the SPANS!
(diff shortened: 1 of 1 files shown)


code/modules/scripting/Implementations/Telecomms.dm
```diff
@@ -244,7 +244,7 @@ var/list/allowed_custom_spans = list(SPAN_ROBOT,SPAN_YELL,SPAN_ITALICS,SPAN_SANS
 		signal.data["verb_yell"]	= interpreter.GetCleanVar("$yell")
 		signal.data["verb_exclaim"]	= interpreter.GetCleanVar("$exclaim")
 		var/list/setspans 			= interpreter.GetCleanVar("$filters") //Save the span vector/list to a holder list
-		setspans &= allowed_custom_spans //Prune out any illegal ones. Go ahead, comment this line out. See the horror you can unleash!
+//		setspans &= allowed_custom_spans //Prune out any illegal ones. Go ahead, comment this line out. See the horror you can unleash!  //Release the SPANS!
```
GunHog, line 247: This is such a bad idea that I cannot even begin to describe how bad of an idea this is. You are a monster if you merge this. But, I do not play here, so it is your call :P

Chronitonity:
Lets see if the first solution works before we add this aye?

nalarmg:
This seems like a bad idea.

code/modules/scripting/Implementations/Telecomms.dm
```diff
@@ -244,7 +244,7 @@ var/list/allowed_custom_spans = list(SPAN_ROBOT,SPAN_YELL,SPAN_ITALICS,SPAN_SANS
...
 		if(islist(setspans)) //Previous comment block was right. Players cannot be trusted with ANYTHING. At all. Ever.
```
GunHog, line 248: You may wish to move this line ABOVE the illegal spans checker, else you may have runtimes for users that do not properly put the spans in vectors. May also need a check for non-string values in the list. I forgot to do this for /tg/, but it does not matter now, as NTSL was removed from /tg/ because MrPerson hates anything fun.

AliiTea, line 248: @GunHog why not join the funbus at hippie then?

AliiTea:
Spans work I am a cuck

blobbernaut:
what would this have done

PR #1002, opened 2015-10-22
Lowers the tech and material value for some implants, RD acccess change

Chronitonity:
Most of these implants aren't even known about, and they never get made due to no one getting any of the proper research requirements (Why does something that ends unconsciousness need syndicate tech??). This hopefully fixes that, and while still keeps them rare, makes it so no one.

It also gives the RD mining access. While I originally wanted to give science some starter materials in order for them to actually DO THEIR JOB, Foos argued that this was a better solution and I'm in agreement. While a good counterplay to this change is "Ask the fucking HoP", its pretty shit that the RD can't go do this himself and help his whole department, all he is right now is just a scientist with robotics and bridge access.

Worst case scenario this is all shit and science does their job and we revert, but who expects that to happen?


Foosili:
Sounds good :+1:

PR #1006, opened 2015-10-23
Selectable equipment update (for admins) and SABER icons fix

Weisyl:
Ports the New-Russian soldier selectable option from pre-rebase Hippie
code. Also adds an ION Corp. Private Military Contractor to the list of
selectable equipment. This can be thought of as a weaker DS-type guy who
isn't endorsed or even really liked by NanoTrasen. Note that ION is not
actually a reference to Dean Hall's game but instead a reference to ION
Incorporated seen in ArmA 2. This would include a polished up CentCom
except I forgot to save an hour's worth of work from yesterday so RIP.
Finally, fixes the SABER / Prototype SMG's icons which were mis-named,
causing the gun and magazines to be invisible.


Weisyl:
Again, PMCs are just a selectable equipment option which is meant to be a substitute for a deathsquad that is still quite capable. For those curious as to what the paper that the PMCs spawn with says: http://puu.sh/kUj8M/86b8a231c5.png

danhibiki:
Why not

Lawligagger:
Looks good to me

PR #1019, opened 2015-10-24
Possibly fixes #1018 , needs testing for live

### LINKED ISSUE #1018: When putting someone in a deep fryer the action doesn't cancel if the person moves away.
You can be halfway across the station and get teleported to the deep fryer. It's really lame when it happens!

TDcoolguy300:
Fixes syntax issues for returns, might of been the cause of the issue.


TDcoolguy300:
Actually, DNM for now. I think i see the issue

Carbonhell:
the returns there should be fine, it's like that in a lot of places,huh, about src i am not sure if that's the right fix

Carbonhell:
Yep, changing src in F fixes the bug. but anyway i prefer the if(x) return layout

PR #1024, opened 2015-10-24
Performance tweaks

Chronitonity:
Fire Alarms no longer trigger every tick

set_airlock_overlays only update when they change

Spiders no longer lock up the Master Controller with do_afters


TDcoolguy300:
:+1:

code/game/machinery/doors/airlock.dm
```diff
@@ -312,13 +319,31 @@ About the new airlock wires panel:
...
 				panel_overlay = get_airlock_overlay("panel_opening", overlays_file)
 
-	overlays.Cut()
-	overlays += frame_overlay
-	overlays += filling_overlay
-	overlays += lights_overlay
-	overlays += panel_overlay
-	overlays += weld_overlay
-	overlays += sparks_overlay
+	//doesn't use overlays.Cut() for performance reasons
```
nalarmg, line 322: You're certain that the below is actually faster/more efficient than using Cut()?

Chronitonity, line 322: About 100% sure of it.

nalarmg, line 322: All my cringe. But merging.

PR #1026, opened 2015-10-24
Adds blood splatter effect pre-rebase,yay

Carbonhell:
that's it,really


Carbonhell:
Added logging to rodguns and fixed a lil error with pindown in turfs

PR #1035, opened 2015-10-25
Reverts a few remaining /tg/ merge armor nerfs

blobbernaut:
A couple of armor items from /tg/ weren't reverted, yet, this should be everything.


code/modules/clothing/head/jobs.dm
```diff
@@ -68,9 +68,16 @@
 	name = "head of security cap"
 	desc = "The robust standard-issue cap of the Head of Security. For showing the officers who's in charge."
 	icon_state = "hoscap"
-	armor = list(melee = 40, bullet = 30, laser = 25, energy = 10, bomb = 25, bio = 10, rad = 0)
+	armor = list(melee = 80, bullet = 60, laser = 50, energy = 10, bomb = 25, bio = 10, rad = 0)
 	strip_delay = 80
 
+///obj/item/clothing/head/HoS/dermal
```
AliiTea, line 74: why not actually add the item in instead of just commenting it out?

This was the first thing I ever did back on old code so >:I

Chronitonity, line 74: Dermal should probably just be recoded to work like an implant, rather than a helmet slot item

nalarmg:
Needs changelog.

danhibiki:
Sounds good to me, just get that changelog in there.

blobbernaut:
changelog has been addededed

PR #1056, opened 2015-10-27
Gibber Fix

AliiTea:
Carbon told me to do it :O


nalarmg:
Needs a better description on what this is doing and why.

Carbonhell:
By default, if a mob gets gibbed it will spit out normal meat. But someone decided to put meat = null to IPCs, therefore overriding the default meat type. When the gibber tries to create the meat type, it tries to create meat(null) and fails,therefore runtimes and the proc stops working. This will make IPCs spit normal default human meat, since it'd be an huge pain in the ass to make them spit metal, which would cause snowflakey code just for a little effect to a race in my opinion useless,ay

nalarmg:
Ah. Please keep descriptions actually descriptive in the future.

Carbonhell:
Yeah well i told what it did in coderchat, sorry ay

PR #1064, opened 2015-10-29
Sec power fixes

Pyko1:
Reduced the amount of lamps in perma/armory, those things are powerhogs and areas are still well lit with less lamps.
Buffed the steam generator to make at least enough power to keep the lamps on at the perma.


Lawligagger:
Looks good to me.

Pyko1:
still removing few more lamps and possibly splitting the sec hallway to sec hallway area and sec lobby areas.

Pyko1:
Removed some more lamps, split the sec lobby in to its own area.
Fixed some Roundstart AT's at eva maintenance airlock.
The remaining  roundstart AT's seem to be caused by the tubetransit, i think.

Pyko1:
Slightly reduced the buff on steam generator.

Its max output is now 2400W, while perma uses 1529W at idle.
In comparison the P.A.C.M.A.N.-type portable generator has max output of 40000W.

Pyko1:
Moved things slightly around in armory. Nothing added, nothing removed.
capture

PR #1065, opened 2015-10-29
Map Fix

Pyko1:
Fixed accidental space tile in insanity ward.
Fixed Isolation APC placement.
Fixed Isolation Airalarm placement.


Pyko1:
Minor area fix at sec. 
Faded photo removed from z4, was causing runtimes according to carbonhell.

Pyko1:
the space tile fix is quite urgent, and all these are fixes. Would be nice if we could get quick merge/compile.

tgstation.dme
```diff
@@ -6,6 +6,77 @@
 
 // BEGIN_FILE_DIR
 #define FILE_DIR .
+#define FILE_DIR "code"
```
Carbonhell: Remove the file dirs

Nebyab:
yeah sure

PR #1081, opened 2015-10-31
Starlight

AliiTea:
Just lowered the values for starlight
thx mso :p

New - 
image
image

Old -
image


AliiTea:
@danhibiki @Chronitonity @Nebyab  merge plox sexy starlight better than less sexy starlight

PR #1083, opened 2015-10-31
SFX for unloading / loading magazines into guns and chambering.

Weisyl:
Every gun that uses magazines have been given their very own sound
effects which play when the magazine is loaded into or unloaded out of
the gun! In addition, chambering (removing) a round from the gun with no
magazine in also plays its own sound.
Shotguns which load via a tube now play a SFX when a shell is loaded.
The total size of the sound folder which adds all these new SFX is 199
KB. Credits to Insurgency (standalone) and Receiver for the sound
effects.


Weisyl:
Here's the folder full of the new sounds for this: https://www.dropbox.com/sh/er9najfvxmx7jay/AADEzaa0oV7hf_Z-eVg8Odl1a?dl=0

Lawligagger:
Yes please

danhibiki:
Alright

PR #1078, opened 2015-10-30
LubeOverlay

AliiTea:
Makes lube use the water overlay.

Credit to PKPenguin321
https://github.com/tgstation/-tg-station/pull/12769

PR #1090, opened 2015-10-31
Fixes error in #1078

Carbonhell:


PKPenguin321:
yeah my bad
i was gonna shout this out to you guys but found out you already solved the issue 12~ hours before i did
i changed it to     if(wet_setting != TURF_DRY) but your fix should work too

AliiTea:
;-; I still can't believe I didn't notice that it wasn't right since I do the exact same thing in NTSL all the time

PR #1094, opened 2015-10-31
Shuttlesound quickfix

AliiTea:
Shuttle now leaves when sound ends.


AliiTea:
@danhibiki Mergorino sound doesn't fit with delay

Chronitonity:
this is a fix

PR #1095, opened 2015-10-31
Map Changes

AliiTea:
Adds cow to escape
Fixes the T pipe in atmos that shouldn't be a T pipe

image The pipe in the middle of this is the T pipe ingame
image


AliiTea:
@Chronitonity honk what about this boi

Chronitonity:
The cow is technically an addition, i'll just wait for someone else

PR #1105, opened 2015-11-01
Updates and fixes to teeth and grabs.

Crystalwarrior:
- Teeth are now stackable items!
- Now the lisp depends on the amount of missing teeth. Examine text tells you how much teeth are you missing.
- Added RIP AND TEAR sound effect to removing embedding items.
- The amount of teeth knocking out now depends on weapon's force. You won't be able to >punch< any more than ~7 teeth now unless you're hulk or something. Punching teeth out, however, is a better idea than to try to use a lower-brute weapon on them. Instead, go grab a toolbox or heck a fireaxe and apply it directly to the mouth for the biggest possible chance to knock dem teeth out.
- Fixed resisting grabs. You can now properly resist out of all grab types (except succesful choking - you can't save yourself now)


danhibiki:
Bug fix really, quick merging

PR #1111, opened 2015-11-02
Teeth necklaces!

Crystalwarrior:
- Teeth necklaces! Now you can wear a necklace with all of your adversaries' teeth! Teeth necklaces can be created by applying cable coil with 4 cables on a tooth. You can hang up to 16 teeth onto the necklace.
  - Multiple teeth types now supported! Humans, Lizards and Tajarans now have unique teeth types.
  - You can now have teeth of multiple types in your mouth! Enjoy having tajaran-lizard-human tooth-filled mouth!


lcass:
The real question is does superfarting knock your teeth out?

PR #1112, opened 2015-11-02
Deepfry overlays fixed for deepfried people

Crystalwarrior:
Deepfried clothes will now properly show up on the avatar if that person was stuffed into a deepfryer.


Crystalwarrior:
Note to self: may have to cache the deepfried overlays for items later since Icon procs can be kind of expensive. Thankfully deepfrying people is not too common and there doesn't seem to be any visible issues yet.

PR #1119, opened 2015-11-03
Spiked Baseball Bats

Chronitonity:
Adds Spiked Baseball bats as a crafting option. They're crafted with one wooden baseball bat and 4 metal rods.

Spiked bats do slightly more damage than wooden bats, but less damage then a metal rod. However they have one nice buff, which is that they can pierce through some (but not all!) of securities armor. This makes them better weapons for tackling security forces.

http://puu.sh/l7s5W/3aeee0a709.png

Includes Inhand sprites, Belt sprites, and ground sprites, all by Oprayx.


danhibiki:
:+1:

Foosili:
Rad.

Nebyab:
yeah syre

PR #1136, opened 2015-11-04
T O R T U R E (wirecutters on mouth)

Crystalwarrior:
Applying wirecutters on someone's mouth will tear out their tooth with small amount of randomised brute and force-scream. Novelty torture feature.
Crowbar suicide_act updated to be much more brutal.
Suicide.dm had another check for restraints added after the suicide_act value was returned.


nalarmg:
Description is not useful or informative.

Crystalwarrior:
Updated description

Foosili:
I dig it. Any way for traitors to do more shit like this is p neat in my books.

PR #1146, opened 2015-11-07
Adds the power fist for traitors

Weisyl:
The powerfist is available to any traitor for 14 TC. When wielded in hand, the user can smash any target across a room with moderate damage and some stamina damage. Needs to charge automatically in-between punches, but the time needed to charge is negligible. Does NOT go in your gloves slot, instead, it must be held in hand like any generic melee weapon. Uses placeholder sprites for the moment, but they're okay for now,


danhibiki:
I think its definitely worth a spin

Chronitonity:
You should probably change all the descriptions referring to it being two handed now

Nebyab:
Yeah sure

Weisyl:
Mergable now since 2 heads said yes?

TDcoolguy300:
Yes, two headmins have approved this :+1:

PR #1150, opened 2015-11-07
Re-adds Antag OOC

Chronitonity:
Re-adds Antag OOC that Whitefox originally ported from Paradise

God is dead and we killed him


Lawligagger:
Sounds good to me.

Nebyab:
I don't recall this ever being used to any significant extent, but its not like it hurts to have.

Yeah sure

PR #1165, opened 2015-11-08
Ejected casing and revolver (un)loading sounds, L6 reload sounds fix

Weisyl:
Ammo casings now make a little sound when they fall after being fired from a gun.
Added sounds for loading or unloading the revolver."
All sounds related to reloading now play properly for the L6 SAW and Abzats.


Lawligagger:
Sure.

PR #1171, opened 2015-11-09
Stops carbon smoke from murdering the server instantly

Weisyl:
That is, until a better solution than commenting the whole dirt-spawning from can be found.


Chronitonity:
merging until someone cares enough to make this not destroy the server

Crystalwarrior:
holy shit you lazy plebs I'll fix it legit.

PR #1193, opened 2015-11-13
Removes the ability to create normal Pneumatic Cannons at an Atmoslathe 

Chronitonity:
It can hold twice as much as a ghetto cannon, fire almost twice as many shots before you need to refill the tank, and to top it off, is unarguably easier to create.


Carbonhello:
Can't you change it with a ghetto one?that'd work

Chronitonity:
Theres already an assembly recipe and i'd rather just keep it like that rather than tacking it onto this.

danhibiki:
Yes

PR #1195, opened 2015-11-13
Adds a new flag "INJECT", allows Saline-Glucose Solution to function as a blood substitute

Chronitonity:
and adds a syringe to the poison kit

All TG station ports.

Inject is for chems that shouldn't be eaten but instead injected or have a special affect when injected, and is used by syringes/hyposprays ect. Before this they used injest, which is the same thing pills used.

Saline-Glucose can be used as a blood substitute if hooked up to an IV drip or injected into someone. Actual blood is better obviously, but this can be used in a pinch, or to "water" down blood.

Poison kit should've started with a syringe to use it, its annoying as fuck having this kit and needing to spend a decent amount of time finding one syringe.


Crystalwarrior:
This sounds pretty great actually, :+1:

Carbonhell:
It's kinda stupid that it wasn't in already...Whatever, yeah, needed

danhibiki:
Sounds good

PR #1196, opened 2015-11-13
Removes the ability for miners to purchase aliens 

Chronitonity:
Interesting idea, terrible in practice when they can create a near infinite amount of other antags easily at the start of the round and can basically force change the gamemode from traitors/traitorlings to station vs aliens.


doobiedoo23:
please no, this shit is cool and only like 2 people actually ever do it

AliiTea:
:-1:

Weisyl:
This is the same bullshit as xenobio being able to make unstoppable zombie hordes. I'm all for this.

danhibiki:
:-1:

trifornt-tinfornt:
I hate Xenobio's Zombie Hoardes, but Alien is way more fun than Xenobio's stupid overpowered zombies.
:-1:

ghost:
sacrificng your round and 20tc for aliens is worth it
niche enough too

Nebyab:
I've never really fancied this idea, so I'm fine with it being removed

Lawligagger:
Same here

Crystalwarrior:
I'm sort of torn about this, actually. On one hand there's the chance the miner using the facehugger isn't even going to be the first drone, or that the player picked for alien fucks up royally and dies somewhere, meaning there's an overwhelming risk.
On the other hand, it may turn a traitorling round into an alien round in a second if there's a robust queen around.

Honestly, :-1:. It's a massive risk-reward that isn't even worth it to some players.

nalarmg:
Two heads have given consent, but given Dan's stance, holding off on merging until resolved.

danhibiki:
Heres my problem, its not something thats going to happen often enough to even give a problem with antag rolls. Right now, on a 40 pop server, lets say 5 people are chosen for antag. Thats less than 15% of the server with antag status. For aliens to become a massive, swarming problem that happens every round, there would have to be a miner and they would have to roll antag. Which, with a 40 pop crew, even if we had 3 miners, its a 7% chance they will roll antag.

I don't see the harm, and I haven't seen massive waves of alien rounds as a result. I think removing it is a knee-jerk reaction to a one off event that we aren't hearing about.

PR #1202, opened 2015-11-14
Re-added the short charge time for power fists

Weisyl:
Getting reports that without it, the power fist is pretty much a super mass-genocide glove of ultimate doom.

### DIFF
--- a/code/game/objects/items/weapons/melee/misc.dm
+++ b/code/game/objects/items/weapons/melee/misc.dm
@@ -139,10 +139,10 @@
 	throwforce = 10
 	throw_range = 7
 	w_class = 3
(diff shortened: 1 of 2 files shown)


Chronitonity:
:+1: Its near uncounterable as it is right now, this is a good change.

trifornt-tinfornt:
Personally I liked to think of it as a more expensive upgrade to the ESword that's more fun to use. I'm not sure if a nerf is entirely necessary, but given how bad I've seen some of the murderbone - it would be at least sort of appreciated though entirely unwarranted.

danhibiki:
Lets do it

PR #1205, opened 2015-11-14
Standardises Space and Hardsuit slowdown from 2 to 1

Chronitonity:
Ports Fox McClouds standardization of Space and hardsuit movement speed
No one wears these things indoors because a slowdown of 2 is near crippling, they always just run with them in their hands and equip them as soon as they need them/are getting disarmed.

Fits with our fast paced gamestyle anyway.


danhibiki:
Just to be clear, this makes them one stage slower than normal, instead of 2 stages? So having a hardsuit/exosuit still slows you down, just not as much?

Weisyl:
@danhibiki yep

danhibiki:
Sure

Weisyl:
Thank GOD.

PR #1209, opened 2015-11-14
Made it so using wirecutters on someone's mouth won't cut their restraints instead

Crystalwarrior:
Means that you can torture someone WITHOUT freeing them preemptively.


Carbonhell:
it's just a changelog,wut

Crystalwarrior:
....huh, then this means I already accidentaly pushed the thing to master.

Hold the fuck on okay it's just general weirdness.

PR #1211, opened 2015-11-15
Power fist rework (code optimization, some balancing improvements)

Weisyl:
- tweak: "Power fist code changed greatly to make it perform better to
  cause less lag on the server."
- tweak: "Power fist cooldown has been readded BUT it uses a different
  system so that rather than having a random 0 - 2 second cooldown which
  was very unreliable you now only have a slightly-longer-than-normal
  cooldown which is practically always the same. New cool-down system also
  lets admins set the cooldown to literally 0 so they can pulverize
  someone as fast as they can click."
- tweak: "Power fist now takes armor into account and no longer does
  stamina damage but can punch through any kind of window or grille
  instantly. Power fist cost has NOT been changed."

### DIFF
--- a/code/game/objects/items/weapons/melee/misc.dm
+++ b/code/game/objects/items/weapons/melee/misc.dm
@@ -128,49 +128,60 @@
 	add_fingerprint(user)
 
 //POWER FIST
-//This thing is a mish-mash of singularity hammer and mjollnir code, but it seems to work well.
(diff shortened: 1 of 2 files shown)


Weisyl:
The bug fix is that the power fist on the server right now ignores ALL armor, which isn't exactly an intended feature.

Chronitonity:
Gonna merge this as it fixes a bug and improves performance.

PR #1217, opened 2015-11-16
New Big Papa's zoot suit, zoot hat, and pimp stick

McHatters:
Big Papa's full ensemble is now available in the Autodrobe! The pimp stick has the same stats as the cane, with just a bit more force added, being diamond-tipped.
First time adding stuff in, hopefully I got this right.


McHatters:
wait, gonna try to fix this again. Something went a bit fucky with my github.

PR #1221, opened 2015-11-16
Removes the bullshit "can't heal in space" mechanic

Crystalwarrior:
Worst part? This only applies to bruise packs and ointments.
Brute patches, etc. aren't affected at all, including the bandages.

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -42,9 +42,9 @@
 				user << "<span class='warning'>[H] isn't bleeding!</span>"
 				return
 
-	if(isliving(M))
-		if(!M.can_inject(user, 1))
-			return
+	// if(isliving(M))
+	// 	if(!M.can_inject(user, 1)) //This doesn't make a lick of sense in gameplay prespective and most other medical tools don't do a check for space suits/etc. to heal (patches, gauzes, etc.)
+	// 		return
 
 	if(user)
 		if (M != user)
--- a/html/changelogs/Crystalwarrior160 - no_bullshit.yml
+++ b/html/changelogs/Crystalwarrior160 - no_bullshit.yml
@@ -1,0 +1,6 @@
+author: Crystalwarrior160
+
+delete-after: True
+
+changes: 
+  - tweak: "Bruise packs/ointments no longer require your flesh to be 'exposed' to be applied. This means you can heal inside a space suit again."


code/game/objects/items/stacks/medical.dm
```diff
@@ -42,9 +42,9 @@
 				user << "<span class='warning'>[H] isn't bleeding!</span>"
 				return
 
-	if(isliving(M))
-		if(!M.can_inject(user, 1))
-			return
+	// if(isliving(M))
```
danhibiki, line 45: Just a quick question, this isn't the code that stops syringes shot from syringe guns from penetrating suits, is it?

Crystalwarrior, line 45: It's the same check used for syringes and stuff, yes.
I'm only removing it's proccall in the medical.dm where bruisepacks/ointments are located.

Lawligagger:
Oh right, this makes it so you can actually use bruise packs on yourself in space right?

Chronitonity:
No point in commenting out code, just remove it.

Chronitonity:
Wait wait wait are you making it so you can't heal through space suits? 

:-1: They're thick for a reason, if someone else can't penetrate your space/hardsuit with a syringe, it doesn't make sense that you can apply a bruise patch or stab yourself with a syringe.

Weisyl:
How come we can use the totally-not-already-overpowered chemical patches with a space suit on though? Plus if I have to do something like bandage myself in space and I'm already badly hurt then I'm literally fucked either way because if I take off my space suit, then space will kill me, if I don't, then bleeding will kill me.

Chronitonity:
Mmm, if you can use patches in space then i dunno why you cant use brute packs, that seems like an oversight.

I'd still prefer if syringes don't go through space/hardsuits if you apply them yourself. That'd just be really weird if you couldn't get shot by a syringe but you could just push it into yourself.

Weisyl:
You could probably push a syringe into yourself a whole lot more forcefully than a syringe fired from a syringe gun several feet away can.

Chronitonity:
Doubtful, I could argue realism but the point stands if that blocks an antagonistic syringe, it should block a syringe from you too

danhibiki:
Waiting for more input

danhibiki:
Also, I don't like the idea of just deleting code for this, incase this is something we want to revert/review later. Just comment it out, even though it will cause (slight) clutter.

Just so we are clear, this is how I'm intepretting it right now. Currently, packs can't be used in space, but patches and syringes can. This will remove the check for a suit on pack usage, which is already present on both syringes AND patches?

danhibiki:
this is a shitpost test
>a
>>b
>>>c
>>d
>e

PR #1235, opened 2015-11-18
Fixes bug with Chemistry update, actually does the changelog

Chronitonity:
i blame foos


Chronitonity:
This should probably get a quickmerge

PR #1246, opened 2015-11-20
Hub forums link update

nalarmg:
The hub forums link now uses the forum URL, rather than being hardcoded within the anchor tag.


Nebyab:
This is important

PR #1249, opened 2015-11-20
Removes Pinning/Forcing people down 

Chronitonity:
Pinning is functionally a permastun for anyone, it requires no equipment to disarm and handgrab someone, stopping them from getting up unless they win a RNG dice roll to get out of it (Which is only 20%). Its super easy to abuse, I ran into a guy who was entering the dorms, pushed him over, grabbed his hands and pinned him, full stripped him, and found he was a cultist, 5 minutes into the round.

As of now it's more powerful than tabling and as stated before, requires nothing besides your bare hands and an iota of skill to change intents moderately fast. It's not something we should balance our entire combat system as giving it to everyone is a flawed concept in my opinion.

I'd be open to doing what Goofball originally wanted to do over at TG, and give Security their own martial art that was good at disarming assailants that includes pinning, perhaps as a way to encourage people to roll security and give them a solid buff that they desperately need, but I don't think we should keep pinning as something every greyshit can use.


Crystalwarrior:
Pindowns in general are already much more ineffective against security members since disarm duration for people with armor is greatly reduced. That also includes table pushing.

I feel that removing pinning down without at least TRYING to balance the GRABBING system, not "entire combat system" like you said is a wrong way to go about this.
I'm planning to make it so instead of reinforcing grab on top of someone you'd have to click on them with disarm intent and the grab itself, an extra step that would give the victim enough time to get up.
For security members, the disarm may last anywhere from 0.1 to 2 seconds because of the way our tick-based stuns work.
For everyone else it can range from 2 to 4 seconds.

We have to make pinning people down either have less window of time or make it much easier to resist, perhaps increasing the chance up to 40%. Not to mention there's an issue with current resisting-out-of-pindown where both assailant and victim get stunned for a similar amount of time which I'm planning to fix.

I'd rather you let me nerf the goddamn thing instead of removing it for a probably permament duration.

Chronitonity:
I don't think it deserves to be nerfed because I don't think it should be here at all, adding a way to stun someone without ANY tools is just a broken idea in the first place. Balancing a grab around an RNG resist is also terrible.

It's a massive (meme incoming) powercreeeeep, that any guy without any tools can stun someone, apply cuffs, or stun someone and just full strip without needing tools. The only other stun that was anything like this was tabling, and thats gone now because it was terrible, and even that needed atleast tables, this just needs your hands and a reaction time of a normal human being.

So no I'd rather not nerf the ability when I don't think it works at all. The concept is flawed, the execution is flawed, its no good.

Crystalwarrior:
I'd honestly rather wait until I hear everyone else's opinion about this before it's removed for good.

Chronitonity:
That is what PR's are here for, so everyone who wants to can input on them.

FuryMcFlurry:
it being locked to security/captain seems like a good idea.

Weisyl:
I agree 100% with Chron here.

Carbonhell:
Removing it entirely feels shitty,it's a nice feature, i'd rather prefer to see resist cooldown being lowered for it, or the chance to resist out be improved. Giving it only to security doesn't look like the best option but it can be tested out,i guess.

Chronitonity:
Putting this on hold (tm) as me and crystal are drawing up some alternatives involving security getting their own style of combat to make them have a bit more teeth, as they are harmless right now.

PR #1250, opened 2015-11-20
Radiation Gun

ghost:
Right now it's only an admin-spawn only for testing purposes. 

New traitor weapon that deals 50 radiation damage (CLONE) and has a
cooldown of 30 seconds per shot. Geneticist traitor-specific item for
about 14 telecrystals seems reasonable.


ghost:
closing for now since i fucked it up

PR #1259, opened 2015-11-21
Chemtak log and explosion

Carbonhell:
- tweak: "Now tanks detonate when exposed to fire/blob/shot at/exposed to an explosion."
- rscadd: "Adds proper logging for chemtank detonating."
  logs=JMP button to the message "triggered a chemtank explosion" and entry in user's attack log.


danhibiki:
chemtak.
:+1:

Nebyab:
yeahs ure

PR #1263, opened 2015-11-22
Holodeck martial arts + Balance Tweaks, fixes

Crystalwarrior:
Adds CQC, Krav Maga and Wrestling arenas to the holodeck for non-lethal practice
Recodes the way disarms work for Krav Maga and CQC AGAIN, so now it prioritizes disarming in-hand item first and if there's no item in-hand it would check the other hand.
Made it so when your permament martial art changes you would "forget" it like you're supposed to
Made Wrestling Belt ignore armor for stuns so both you and your opponent are stunned for the same amount of ticks. Armor is still considered for applied damage.


danhibiki:
:+1:

Nebyab:
sounds good

PR #1267, opened 2015-11-24
New cat grenade, fedora tipping action button

Weisyl:
You can now use an action button to tip a fedora if you're wearing this.
Doing this too much is not recommended by your local Chief Medical
Officer. Detective's fedora is also a subtype of the regular fedora now.
Added a grenade for traitors that spawns 8 feral cats when detonated.
Costs 5 TC.


danhibiki:
Sure

Foosili:
Question about the cat grenade, what really makes this different from the viscerator grenade? How much health do the cats have? What kind of damage do they do?

Weisyl:
@foos 5-10 brute per attack (varies from 5 - 10), 30 HP per cat

Lawligagger:
Sure.

PR #1275, opened 2015-11-25
Rotatium

TDcoolguy300:
Adds Rotatium, a chemical that causes the person who consumes it to rotate their view, causing distorted shit like the proc would do.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Medicine-Reagents.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Medicine-Reagents.dm
@@ -346,6 +346,7 @@
 	for(var/datum/reagent/R in M.reagents.reagent_list)
 		if(R != src)
 			M.reagents.remove_reagent(R.id,2.5)
+			M.client.dir = 1
 	if(M.health > 20)
 		M.adjustToxLoss(2.5*REM)
 	..()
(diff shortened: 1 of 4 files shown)


TDcoolguy300:
Also, made with Teslium, Mindbreaker, and Spacedrugs-Neurotoxin(Not the drink)

Chronitonity:
chaaaaangelog

TDcoolguy300:
Changelog done

Chronitonity:
May want to add that Calomel purges the effect from you

TDcoolguy300:
Done

ghost:
is the shooting and shit still bugged, this stuff crashed my byond when the rotation stuff was happening

Crystalwarrior:
:-1: if crashes/weird shit aren't fixed yet
I'd rather have the view rotated in such a way that all sprites face the correct way

TDcoolguy300:
Then this will never exist then Crystal, this rotate's the client in a way that Autosmoothing will not work for the effected client. This will not cause server crashes either because this is effect is entirely client based.

Crystalwarrior:
I'm talking about client crashes which already have been reported.
I seriously don't think this weird visual effect that fucks with the client's perception and destroys his view in a way that can even cause a client crash is worth it in this case.

Nebyab:
As long as it doesn't crash people I'm good for this even though the rotation really bothers me

danhibiki:
We can do a test run of it, but if we have mass reports of it causing crashing, we're gonna have to back out.

TDcoolguy300:
I'm cool with that one, this is easy to revert as long as nobody fucks with the dms too much

PR #1278, opened 2015-11-25
Re-added megascreaming

Weisyl:
Re-added megascreaming. Uses new sounds instead of just being the
regular scream sound spammed half a dozen times. Currently uses 2 sounds
and is only available for males.
Sounds:
https://www.dropbox.com/s/dgzjaeuvm7f1ahb/scream_male_mega1.ogg?dl=0
https://www.dropbox.com/s/jhbophd1sfkqg8x/scream_male_mega2.ogg?dl=0

If this passes...wew lad.

PR #1284, opened 2015-11-25
Revert "Re-added megascreaming"

danhibiki:
Reverts HippieStationCode/HippieStation13#1278


TDcoolguy300:
Rip Megascream

danhibiki:
Lets just wait a bit before we commit, and let it get discussed if we really want it or not.

PR #1285, opened 2015-11-25
Prevents accidental IC in OOC

doobiedoo23:
OOC will catch messages that start with "." ";", ":", "#" (the say channel modifiers) or "say" itself.

When you type something that looks like IC in OOC you'll get a pop up that will let you actually say it if that's what you meant to do. It defaults to no so real accidental IC OOC can mash out of the command in case what they said was something like ";help I'm being killed!"

Credit to incoming5643 on /tg/ for this


Chronitonity:
:+1:

Nebyab:
yeah sure

AliiTea:
if only we could remove the ooc hotkey.. or better yet have it so when you hit send it asks if you want to send it

PR #1288, opened 2015-11-26
Fixes char setup screen

Chronitonity:
from tg because im an idiot. stops things like implants or the captain coming to the station being announced when you put them to high. this should be quick merged.


danhibiki:
:clap:

PR #1300, opened 2015-11-27
Embedding fix

TDcoolguy300:
During my tests nothing ended up staying in the bodies, so this SHOULD work for fixing infinite embedding


Carbonhell:
Did you manage to actually reproduce the bug before applying that patch?i wasn't able to

TDcoolguy300:
I've tested it under what i have seen in the server, where the embedded object stays in the limbs of the subject and not the main object.

TDcoolguy300:
Also, i've tested this and after embedding someone in the limbs, the removal of said embedded object worked fine

PR #1302, opened 2015-11-27
Makes Anomalies more dangerous, makes have a chance trigger more often

Chronitonity:
Ports Anonus's changes to Anomalies from TG 

Anomalies move more often and are more resistant to explosions
Gravitational ones should throw objects at mobs
Hyper-energetic flux ones shock mobs that get near or run into it
Vortex ones will sometimes throw objects at nearby living mobs
Vortex anomalies will sometimes throw objects at nearby living mobs
The Bluespace anomaly will teleport mobs away from it now
Fire anomalies will produce better fires and will release an additional burst if not stopped.

Events should now trigger every 3-10 minutes instead of 5-15. Events barely happen as it is, and they spruce up the round nicely.

The False Alarm event will only pick from events that warn the crew now, fixing the fact that they could trigger and let no one know that nothing happened.


danhibiki:
Sure

calibraptor:
Yes please

PR #1304, opened 2015-11-27
Projectile bugfixes + screen flash no longer prevents all interaction

Crystalwarrior:
- tweak: "Makes screen flash mouse-opaque so you can still shoot/interact with objects while flashed. Solves the issue where you literally couldn't shoot or defend yourself in any way when flashed."
- bugfix: "Overhaul to projectile.dm - made projectiles density 0, reworked Bump proc, simplified range procs. This means that projectiles won't randomly miss you in most cases where they did, and means projectiles won't randomly hit you when they shouldn't. Thanks to /tg/ for most of these.

### DIFF
--- a/code/_onclick/hud/alien.dm
+++ b/code/_onclick/hud/alien.dm
@@ -142,6 +142,7 @@
 	mymob.flash.name = "flash"
 	mymob.flash.screen_loc = "WEST,SOUTH to EAST,NORTH"
 	mymob.flash.layer = 17
+	mymob.flash.mouse_opacity = 0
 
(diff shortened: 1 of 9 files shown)


Chronitonity:
Dunno on the flash change, its a significant nerf

Crystalwarrior:
You gotta admit, though, that it's frustrating as shit when the only thing you can do when flashed is run around like a headless chicken.

ESPECIALLY since it prevents you from shooting, hitting things, or doing anything period AND it's a fucking Area of Effect.

trifornt-tinfornt:
Agree, pretty damn annoying.

danhibiki:
Sure

PR #1312, opened 2015-11-28
Adds two new balaclavas

doobiedoo23:
This PR adds two new balaclavas; the respirator and the three-holed balaclava. The respirator (based on the one from MGO) can be used like a gasmask. Both of the new masks can be found in the Clothesmate, and respirators can be found in engineering lockers. The sprites aren't amazing but they can probably be updated at some point, i just fuckin love balaclavas and think we need more


doobiedoo23:
balaclavaground

The half-balaclava wasn't added due to it being too similar to a bandana

Lawligagger:
Sure

PR #1325, opened 2015-11-29
Allows different tools to deconstruct/construct things faster

Chronitonity:
Ports TG's toolspeed var.

Currently this only effects bigger crowbars, the syndicate screwdriver, and experimental welding tools

Ups force of wrench and crowbar to 7.


Crystalwarrior:
Ooo, this sounds coolio

danhibiki:
sure

danhibiki:
Only thing I'm concerned about actually is we are porting TG values, which means its going to be drastically slower.

Nebyab:
see,ms good

PR #1395, opened 2015-12-08
Brig Fixes

Kilmannan:
Er. Check this, because I've not a clue what I've done with Git. This should be a few small tweaks to the Brig, including a Security corridor from Muster into the Brig.

There is every chance this is not that, and instead a total cluster of fucks.


AliiTea:
Screenshots?

Kilmannan:
I have no idea what I'm doing with this. I've committed updates which should have fixed the mistake I made where the last PR didn't actually have any changes. I don't know if it's registered here, and you can pick and choose or what.

image

TrashFiend:
Does this fix certain doors being uncloseable due to firelocks blocking them? Mainly all the ones connected to the lobby.

Nebyab:
i dunno sure

ghost:
this most definitely fixes the firelocks being ontop of the normal, closed airlocks
right

PR #1399, opened 2015-12-08
Reimplements cat bans

nalarmg:
Cat bans are now a type of jobban. Not the best method, but it saves reinventing the wheel.
Cat banned players can only play as assistants.
Cat banned players (due to the Tarajans punishment role) will not be eligible to be made into an antagonist. This includes conversions and shadowling enthralling.


Chronitonity:
dont remove cats from roundstart, people like to play them to fuck about

Carbonhell:
Code seems good to go +1

danhibiki:
I'll agree with this for now, we can always make people cats.

Nebyab:
I think we need to have a discussion about catbeasts. I think cats are a punishment role, and the catban does thusly serve a purpose, but, I also know that people like to play as cats, so I think we need a roundtable discussion about whether or not we want them to be available to the average player for shits and giggles.

Weisyl:
How about we let people play as catbeasts but also let admins have the option to lock them to it.

Chronitonity:
:+1:

trifornt-tinfornt:
If locked into a cat, you're banned from antag.
If playing as cat, you're not.
Sounds a tid better in my opinion.
aside from that :+1:

SplergeBucket:
I have a rock-hard metaphorical erection.
:+1:

TDcoolguy300:
:+1:

PR #1410, opened 2015-12-09
Reactive Armor Nerf

AliiTea:
Lowers the range at which the reactive armor teleports you from 6 to 4


Weisyl:
YES TAKE MY NON-EXISTENT MONEY

calibraptor:
Nerf Camille

danhibiki:
Yes

Chronitonity:
"Balancing around 1 player is terrible balance"

http://puu.sh/lPSHi/4ac3460173.png

http://puu.sh/lQzsy/225b852b58.png

Stop.

trifornt-tinfornt:
This actually wouldn't nerf the thing much at all, it'd still teleport you away from most of danger and be incredibly annoying and OP.

SplergeBucket:
I'd prefer stuns against reactive armour lasting longer, EMPing the user, perhaps dealing burn damage, etc. than just lowering its teleporting range since the armour really has little in the form of a downside other than hoping they hit space.

Crystalwarrior:
I second what Looping is saying, nerfing teleport range does jack shit in combat situations and it still makes it just as annoying as wizard's blink.
I'd rather either have limb checking so teleport armor ONLY protects from direct hits to the chest, or that it would have a battery charge to limit the amount of uses you can squeeze out of it.

SplergeBucket:
It's just a very situational negative for this armour, the teleportation. 
Hell, if anything lowering the teleport range decreases how situational this negative is, and allows anyone with enough/the right access to still get out and flat. If you truly wanted a nerf you'd give the armour a chance to malfunction like standard teleporters and fling you off into space.

Nebyab:
Not only is this a pointless nerf, but you're literally just targeting one player.

trifornt-tinfornt:
I think that if the teleport armor is EMP'd, it should stun the user and not work for a while as well as having a recharge before it can teleport again. Malfunctioning would also be nice.
Then again, what do I know?

ghost:
how about
add emagging the thing for an even larger range..

danhibiki:
Two nos, one yes.

PR #1413, opened 2015-12-10
Small CQC upgrade - Solves grabs being underused as fuck

Crystalwarrior:
grabs were underused as shit due to their randomness, not to mention the lack of aggressive/blue grab stage move to keep someone down.

This gives CQC users a legsweep move that can be executed with harm act and aggro grab, as well as gives you 100% chance to aggro grab when standing directly behind someone (similar to sleeping carp's back kick)


Chronitonity:
cqc hardly needs an upgrade imo

Crystalwarrior:
I feel that the combo actually kind of detracts from the entire point of CQC, though.
it's really much less stealthy to beat someone up with 5 kicks than to sneak up on them, grab them from behind choke them out.

danhibiki:
Sure

ghost:
gives more incentive to grappling, so this should distract people from the 5  hit meme combo

Nebyab:
We can give it a shot, I'm not sure this is really needed per se, but it seems like it could be an improvement,

PR #1418, opened 2015-12-11
Master

Weisyl:
Woops
Also there's no changelog for this considering that even being able to buy the fatigues was meant to be  a hidden feature until people eventually find out about it and blab.


Chronitonity:
buuug fix

PR #1434, opened 2015-12-13
Arrivals Overhaul

Kilmannan:
Extended arrivals, replaced with a customs style entry system with two lanes, conveyors, shutter control, etc.

Gives the area more revisit value throughout the round, gives Sec a chance to play a function if they want to, and also creates an area that might be of use to well organised antags.

image


nalarmg:
Conflicts. You need to fix the conflict locally and commit the fixed revision.

danhibiki:
+1

Kilmannan:
If only I knew how and what a conflict was........ Anyone?

PR #1440, opened 2015-12-14
MetaOps

AliiTea:
{#} Credit to KorPhaeron
_wew copypasta_

Inspired by HBL's "meta ops"

Syndicate nuke team leaders now have an item that allows them to announce their impending assault on the station, making their shuttle unable to leave for 20 minutes in exchange for a massive telecrystal bonus (the fluff being that by showing off like this you attract more support from various syndicate benefactors).

The item has the following restrictions

-Can't be used if its more than 5 minutes into the round
-Can't be used if you've already left the base
-Can't be used with less than 50 players online

tl;dr the crew gets 20 minutes to prepare for ops but the ops get a ton of extra gear.

Still need to test everything/think over the exact timers and telecrystal bonuses


danhibiki:
Maybe we could look into lowering the wait time for them to jump on ship, but over all :+1:

Weisyl:
I was hoping they'd get access to more gear along with the TC bonus.

trifornt-tinfornt:
"20 Minutes"
Holy god, nobody has the patience to wait that long - extra gear or not. Also chances are by the time the nukies showed up the shuttle would have been called, there would be a Durand or something walking around and generally everything would be chaotic and extra TC or not the nukies would be kinda screwed. To highlight this, let's say the Nukies announced they would attack and they use their extra TC to buy things like energy shields and whatnot. This would backfire immensely because Sec would learn very quickly to raid the armory for Riot Shotguns and chances are a good Roboticist would crank out a mech that can use a machine gun.

I always thought that the Nuke Ops worked best if they attacked out of surprise early round where nobody is expecting a guy in a red hardsuit to start kicking their asses.

I'm okay with the idea but 20 minutes is more than enough time for the crew to get gear to start kicking the Nukies' asses as soon as they step foot in the station unless they get enough telecrystals to afford a bunch of cyborgs or mechs.

danhibiki:
Don't sweat lad, already discussed that 20 minutes is too long. We're talking about moving it down, but too low and we feel we might need to nerf the TC gained. Its still all in progress, but thanks for the detailed input.

ghost:
add different spectrums of the special gear, like lets say metal gear solid's own diamond dogs, with CQC and tranquilizer guns... and some other shit mixed in there

trifornt-tinfornt:
Also I forgot to mention Gateway and how insanely OP that is. Crew can just get Nuke Ops gear themselves if the delay is too long which is so incredibly dumb that I'm surprised you guys said that Gateway is fine.

Lawligagger:
I actually really love the idea of it, I think about 15 to 12 minutes should be a good wait time.

trifornt-tinfornt:
My main gripe is - what would the Nuke Ops do on that wait time? Once they have their gear chances are they're just gonna sit around and hope something happens - like Cayenne suddenly being possessed by an admin.

Chronitonity:
if rhey aren't patient enough to wait 15 minutes (More like 7ish give or take after setting up and buying stuff) AFTER they decided to buy this and get 200 TC then thats on then

danhibiki:
We've discussed that, and its mainly going to be waiting after the planning and buying is done, which sucks, but is the cost of such a high boost in TC. We've also discussed a potential fake challenge item that would produce the challenge message, but at a high TC cost.

SplergeBucket:
12-15 minutes sounds good to me, I reckon since it's a lower wait time that the amount of TCs gained should also be lower. Perhaps at 140-160?

code/game/machinery/computer/syndicate_shuttle.dm
```diff
@@ -1,3 +1,5 @@
+#define SYNDICATE_CHALLENGE_TIMER 6000 //10 minutes
```
Chronitonity: 15 minutes cuckoldia

PR #1442, opened 2015-12-15
{ORIGINAL AUTHOR IS KILM} Arrivals Overhaul

Weisyl:
Complete Arrivals overhaul. This area now has a customs feature whereby Sec can, if they want, man it and search those joining the station late. It just adds a bit more depth to a historically underused area of the station and gives potential for antag use.
The old Arrivals sec checkpoint is now the Invasive Search Area
Sorted out no power to the two APC's in the Brig


Weisyl:
We be doin' this shit Papers Please style now.

Picture:
http://puu.sh/lW2Mk/c079a91223.png

trifornt-tinfornt:
This is kinda dumb, and I guarantee some jackass will close the shutters and do shit that will trap new arrivals.
This would be fine for a roleplay server.

Kilmannan:
Easy enough to circumvent and green lane should start in open position.

SplergeBucket:
Not a fan.
Behold my feedback in the form of this.
ss 2015-12-14 at 08 22 49

I believe it'd be far superior to have a simpler and smaller customs right here
could ve done it better here

Kilmannan:
Any particular reason?

SplergeBucket:
It just feels too unnecessary and big, as my second image is, I reckon you could've done a better job if you focused on re-vamping up that area rather than creating an entirely new one

Kilmannan:
Bizarrely it doesn't feel as big as it looks when you walk around it. It's compact and a fairly decent fit. I did try a version where I cut out and redesigned the waiting area at the current sec point but it felt cramped and seemed like a half-effort. 

In terms of gaining access with the shutters down,  the tables can get moved to the other side to make smashing windows easier.

The old sec checkpoint is a strip search area and would provide antags with another room. I see arrivals customs getting used as gangs / cult base etc.

I think it doesn't do any harm and it'd certainly be worth a try with tweaking as play is observed and issues identified and the flow is worked out.

SplergeBucket:
In all honesty I can't see it being used as anything but a Cult/Gang/etc. outpost. It requires effort on Security's behalf to monitor the region and with little result as late-join antags immediately using their PDA in the arrivals shuttle is rare.

I can see it being used to close off the area by a guy who wants to fuck with people though, which is why I wanted those four tables removed, so as to make it easier for people to break through the outpost if it's shut.

Kilmannan:
Yep, its still bypassable as it is, and will look at making it more so. Would you be against giving it a trial period? I generally sit as ghost and watch for a few rounds to see the flow of movement and interactions.

SplergeBucket:
I've got nothing against a trial period. It's good to let the players develop an opinion about something we're considering here.

PR #1443, opened 2015-12-15
Chaplain Holy Armor

Chronitonity:
Ports TG's fixing of glowshrooms and allowing items in hand to slowdown people

Also adds the tesla armor but reworks it to be a chaplain specific item. 50% of the time it blocks attacks to the chest and shocks people within 3 tiles. If you don't want to get shocked, aim somewhere else. As a trade off it has no armor stats at all compared to the normal templar armor (40 brute defense.)

It costs 14 TC.


danhibiki:
failed chekz.

But yes

Weisyl:
TBH I think the chance is too high. Maybe 25% to 30% chance to shock people instead?

SplergeBucket:
Sounds good to me.

Crystalwarrior:
@DerptheStewpidGoat then it'd be completely useless and you'd be better off just stealing reactive teleport armor
Something like that for 14 TC is good imo.

FuryMcFlurry:
Maybe give some other sprites?

McHatters:
What sprite would this use anyways?

FuryMcFlurry:
the shitty gimmick armor he gotten. so same sprites

McHatters:
so indistinguishable from the armor that spawns in the chaplain's closet?

FuryMcFlurry:
Never played chaplain, but the files indicate for the templar armor.

Chronitonity:
they're the same on purpose.

PR #1452, opened 2015-12-16
Arrivals tweaks

Kilmannan:
ThiA quick fix to some issues in arrivals including a space tile in the arrivals cockpit an and and the APC's not being wired correctly.


nalarmg:
Requires proper map cleaning/merging.

PR #1456, opened 2015-12-16
Does Nexendias changelog for metaops

Chronitonity:
I love it when my boyfriend puts Vitamin D milk inside of my butthole using the classic Dildo Funnel. Once, he even poured in a cup and a half of Grapenuts as well. Breakfast is served, you know what I'm saying? But we're working on a dinner course now. Maybe try out some mashed taters, and perhaps even brown gravy.


ghost:
aga

AliiTea:
Funny thing is some of that is actually true :^)

PR #1471, opened 2015-12-18
Shuttle Tile tweaks and more

Chronitonity:
First two are ported from TG
Shuttle tiles that have been destroyed now no longer drag stationary space tiles with the shuttle when it moves, they function as actual holes that people and things can fall out of. (Also they're animated too!)

The escape pod safes now contain a bluespace capsule, this functions as a "house in the box" to use in unsafe environments.

Suicide text for the nooses aren't as edgy anymore, fixes a lot of suicides just not working because someone commented them out for some reason lol.


Weisyl:
mfw singularity is actually bad for the escape shuttle now

Foosili:
Thank fuck, good shit :+1:

PR #1494, opened 2015-12-21
Moves Grounding rods into secure storage

Chronitonity:
Makes it super easy to make infinite power elsewise. This way atleast requires a teensy bit of work.

http://puu.sh/m3iPa/22deefb586.png


Iamgoofball:
> didn't map merge

top kek/10

Chronitonity:
man its 3 am i aint got time for that shit

Foosili:
im for it

Iamgoofball:
:-1:
Here's whats gonna happen
Players set up tesla, forget grounding rods
All die from shock bounces 
Round ends from other shit
Players set up singularity instead of tesla and never again touch tesla because it takes new effort and they're too lazy

danhibiki:
I think goof has a point, but I also think we should make this change later, once the engine has more familiarity with the player base.

Foosili:
I think waiting on it for players to be more familiar with the engine is fine but the tesla engine can't be as EASY to use as it is now imo

SplergeBucket:
Maybe a trial period for this after a while.
till then :-1:

danhibiki:
Its just that its a totally new engine, and we can actually compare this to the supermatter. The supermatter saw extremely limited use, and a lot of that was not knowing how to properly operate it. With the rods already being in place, this will stand as a foothold for people to start learning it, rather than becoming immediately turned off

PR #1512, opened 2015-12-24
Nudge.py call fix

nalarmg:
Fixes data being sent through nudge.py sometimes being cut off arbitrarily when sent through slack APIs. Shouldn't have any negative effects on the IRC bot either.


nalarmg:
Worth noting - the bug this fixes also effects the IRC bot. This fix also works for the IRC bot.

TDcoolguy300:
I'm all for it

PR #1520, opened 2015-12-25
Smoothwall Fixes

danhibiki:
Variety of ports from TG's smoothwall code, should help with random away mission visuals and should be faster in some cases


TDcoolguy300:
This good to go Dan?

danhibiki:
yes

PR #1522, opened 2015-12-25
Adminhelp and notice to IRC/Slack rework

nalarmg:
Centralizes adminhelp and admin notice to IRC/Slack handling. 
Separates the process for calculating the admins away, online, and without ban powers.

Rather than separate checks before each attempt to send to IRC/Slack in individual files, all handling is done by send2irc_admin_notice_handler(). On modification, requires an extra switch case added with the type of message being handled and the methodology of handling.

Allows sending adminhelps to IRC/Slack regardless of away status of online admins and adds the option to the initial config.


TDcoolguy300:
I assume this is good to go Nalar?

nalarmg:
Yep.

PR #1524, opened 2015-12-25
More verbose round ending (IRC/Slack).

nalarmg:
Displays the round type, surviving humans and observers when sending a round end notice to Slack/IRC


danhibiki:
:+1: Going to approve and merge

PR #1526, opened 2015-12-25
Specifies what child of singularity is spawned

danhibiki:
So instead of saying singulo for a narsie summon, it will list narsie. Likewise for the tesla.


Nebyab:
sure

PR #1536, opened 2015-12-26
Modular Vehicle framework and content

TDcoolguy300:
As above, this is the Modular Vehicle PR, 
# THIS IS HIGHLY ALPHA, AND VERY DIRTY

more info on the Trello 
https://trello.com/c/w6K52Y9F/677-vehicles


danhibiki:
YEsssssss

AliiTea:
TinyTanks 2.0

TDcoolguy300:
Except they are not a child of MECHS

doobiedoo23:
:+1:

AliiTea:
Hardy har har har make the tank barrel not shite plox

code/modules/vehicles/vehicle.dm
```diff
@@ -0,0 +1,381 @@
...
+				user << "<span class='notice'>You repair some damage to [src]</span>"
+			else
+				user << "<span class='notice'>[src] is at full health</span>"
+		else
+			user << "<span class='notice'>The [WT] must be on!</span>"
+
+	if(istype(I,/obj/item/weapon/reagent_containers/fueltank) || istype(I,/obj/item/vehicle_parts) || istype(I,/obj/item/weapon/stock_parts/cell))
+		var/obj/item/vehicle_parts/O = I
+
+		if(O.part_type == A.part_type || O.part_type == E.part_type || O.part_type == P.part_type || O.part_type == CH.part_type || O.part_type == F.part_type  || O.part_type == C.part_type)
```
TDcoolguy300: This line was refactored into a proc

SplergeBucket:
INTERESTED YES

code/modules/power/cell.dm
```diff
@@ -10,6 +10,8 @@
 	throw_speed = 2
 	throw_range = 5
 	w_class = 2
+	var/part_type = "cell" //I blame vehicles
```
nalarmg: As far as I can tell, this isn't actually being used. Surely it'd be better to do a type check rather than this?

code/modules/power/cell.dm
```diff
@@ -10,6 +10,8 @@
...
+	var/weight = 0 //VEHICLES
```
nalarmg: Also not actually being used. These redundant variable seem to be caused by you trying to convert this object to another. Possibly create a dedicated cell type rather than adding variables that aren't used, if this is the issue?

code/modules/vehicles/vehicle.dm
```diff
@@ -0,0 +1,403 @@
...
+			else
+				usr << "<span class ='warning'>You stop removing [P]</span>"
+			interact(usr)
+
+/obj/vehicle/proc/build_text(obj/item/P)
+	. = ""
+	var/name_text = ""
+
+	name_text ="<A href='?src=\ref[src];parts=\ref[P]'>[P.name]</A>"
+	. = "[name_text]<BR>"
```
nalarmg: Whilst not functionally and theoretically not syntactically incorrect, should be lowercase to remain consistent.

code/modules/vehicles/vehicle.dm
```diff
@@ -0,0 +1,403 @@
...
+	if(!vehicle)
+		return
+	vehicle.exit()
+
+/obj/vehicle/New()
+	for(var/obj/item/O in parts)
+		O.loc = src
+	var/obj/item/vehicle_parts/engine/E = locate() in parts
+	var/obj/item/weapon/reagent_containers/fueltank/F = locate() in parts
+	F.reagents.add_reagent(E.fueltype, F.volume) //Debug stuff
```
nalarmg, line 237: Debug code for something that isn't necessarily being debugged live?

TDcoolguy300, line 237: Thats only there to fill the fuel tank up to max when i spawn the stupid thing, it will be taken out before this is merged.

code/modules/power/cell.dm
TDcoolguy300: compareParts actually needs this, otherwise i had some issues with determining whether there is a cell already installed in the vehicle

nalarmg: What issues exactly? This is not the best way of doing it at all.

TDcoolguy300: It was an attempt to stick all the part adding into a single istype chain, but if you insist on me seperating and removing the variables, i'll go ahead and do so.

nalarmg: I do, creating variables that aren't really being used as variables is not good practice, and incredibly dirty. You could do an istype check and add a bit of extra (admittedly snowflakey) logic rather than doing this.

code/modules/vehicles/vehicle.dm
```diff
@@ -0,0 +1,403 @@
...
+			occupants_max += 1
+		O.weight += mass
+	update_icon()
+
+/obj/vehicle/proc/click_action(atom/target,mob/user,params)
+	if(!driver || user.loc != loc )
+		return
+	if(user.incapacitated())
+		return
+	if(!locate(/turf) in list(target,target.loc)) // Prevents inventory from being drilled
```
nalarmg: Redundant and not needed. Might be needed in the future if drills are planned, otherwise this can probably be removed.

TDcoolguy300: Already removed locally

nalarmg:
Everything I've not noted is probably okay. Recommend going through and double checking for any redundant or not needed code. As this is a framework, care should be taken not to bloat things and keep things neat, clean and (relatively) easy to use.

TDcoolguy300:
The massive vehicle.dm file will be seperated into other DMs to keep things tidy and grouped

TDcoolguy300:
http://puu.sh/maKHU/4bf555945d.gif

code/modules/vehicles/vehicle.dm
```diff
@@ -0,0 +1,328 @@
+/*
+This is the base file for the Artic Station Vehicle code,
+Dont touch this file unless you know your shit about this.
+*/
```
TDcoolguy300, line 7: For this, the original purpose was for a planet-conversion of SS13, currently on hiatus and i might aswell port it here to work on it

TDcoolguy300:
Just going to close this for now, i'll reopen it when it's time

PR #1558, opened 2015-12-29
Nanoui Ports

danhibiki:
Lets see if this will work


SplergeBucket:
wel does it work??

danhibiki:
It does work, but I need to port a lot more fixes to be up to date, and theres no reason to push an incomplete project right now when all of the systems we have now are working. This is massive really, and I'm going to have to do more files again, so it'll probably be a day or two before its ready to roll out.

PR #1581, opened 2016-01-02
Fixes teeth being more heavy than a pack of cigarettes

Kokojos:
I want my pool to absorb those fuckers in the drain.
# poolmasterrace


SplergeBucket:
This is more or less a bugfix. It shouldn't have a weight class of +1.

PR #1586, opened 2016-01-02
Brings back the gun closet in the armory, adds two windows to the sec lobby door

doobiedoo23:
These were two major complaints from the players; Add more vulnerabilities to sec (because they're getting too strong) and bring back the gun closet in the armory because it's integral to game modes such as nuke ops.

I tried not to change too much because I genuinely like most of the sec changes but these seemed like important 'fixes'.

http://i.imgur.com/z337fWN.png

http://i.imgur.com/qsgKtYg.png


doobiedoo23:
OH CHRIST I FUCKED IT

Issue #1575, opened 2016-01-01
Grammatical Error - Loyalty Implant Case


Chaznoodles:
"Cell 1 has been hit with the implant case - 'Loyalty' by Steven Vasilyev to no affect!"

Should be effect, not affect. Git gud grammars.

TDcoolguy300:
Blame TG

PR #1587, opened 2016-01-02
Grammar fix

AliiTea:
Changes "Affect" to "Effect" on the attack message you receive when attacking a glass blastdoor.

Fixes: https://github.com/HippieStationCode/HippieStation13/issues/1575


SplergeBucket:
danke for the fix

PR #1597, opened 2016-01-03
Fixes bug where pAIs could not use messenger or chatroom.

TrashFiend:
This should fix the issue where pAIs cannot use the Digital Messenger or Chatroom Client because they don't have the dexterity. Tested and it seems to work fine.

I was worried this fix would let pAIs control stuff they shouldn't be able to, but they still seem to be limited in the same ways they were before.


TDcoolguy300:
Bug fix, going to merge

PR #1607, opened 2016-01-05
Announces forced secret modes in slack/IRC

nalarmg:
The server will now send an announcement to slack/IRC whenever an admin forces secret to a particular game mode.


danhibiki:
Yes!

PR #1645, opened 2016-01-14
Blobcode

blobbernaut:
Adds changelog. Removes meeseeks boxes as per Admiral Hippie's request (Because it might be causing runtimes on the server, don't panic)


Nokington:
yes

PR #1652, opened 2016-01-15
Defib/cloning pod automatic body re-entry

blobbernaut:
The intention is to allow people to have a preference that lets them choose if they want to AUTOMATICALLY enter their body upon being cloned. 

I am not familiar with any of the procs in question, and can't test locally, so I'm putting this up for people to look at to see if it will do what I intend.

DO NOT MERGE


code/modules/mob/mob.dm
```diff
@@ -917,8 +917,12 @@ var/list/slot_equipment_priority = list( \
...
 		ghost.notify_cloning(message, sound)
 		return ghost
-
-
+		
+/mob/proc/force_reentry()
+	var/mob/dead/observer/ghost = get_ghost()
+	if(ghost)
+		ghost.reenter_corpse()
+		return ghost
```
nalarmg: Not sure as to the purpose of returning the observer mob object if they've re-entered? A better way would to be to have it return either 1, or 0 (success and fail/true and false respectively) if you need to detect whether or not the procedure has completed/failed.

blobbernaut:
Okay, I've tested it by calling the relevant procs individually on a local server. Tentatively read to be merged.

danhibiki:
What about what nalar said about your return?

blobbernaut:
tbh I just copied the "notify_ghost_cloning" proc's format but I actually can't think of a reason.

blobbernaut:
ready to merge

danhibiki:
:+1: from me

code/modules/client/preferences_toggles.dm
```diff
@@ -247,6 +247,15 @@ var/global/list/ghost_orbits = list(GHOST_ORBIT_CIRCLE,GHOST_ORBIT_TRIANGLE,GHOS
...
 	feedback_add_details("admin_verb","ITENTS") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
+	
+/client/verb/toggle_auto_reenter()
+	set name = "Toggle Automatic Re-entry on Revival"
+	set category = "Preferences"
+	set desc = "Choose to automatically re-enter your body when you are cloned or defibbed"
+	prefs.toggles ^= FORCE_REENTER
+	src << "[(prefs.toggles & FORCE_REENTER) ? "You will now receive a notification when you are being revived." : "You will now automatically enter your body when being revived."]"
+	prefs.save_preferences()
+	feedback_add_details("admin_verb","FrEnter") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
```
ghost, line 258: this is not a `admin_verb`

Nokington:
sure

code/modules/client/preferences_toggles.dm
danhibiki, line 258: Its the same format toggle intent style uses, which is also not an admin verb. Unless i've been privileged this whole time.

blobbernaut:
revert, does not actually do anything. Give me a while to figure out how byond works; this is actually not as simple a change as I thought it'd be.

PR #1661, opened 2016-01-17
Fixes bug where AIs can't track people when an NTSL script changes $source

TrashFiend:
Not in the issue tracker, but has been a bug on /tg/ since at least June 2015 (https://github.com/tgstation/-tg-station/issues/10032).

If someone uploads an NTSL script that changes $source, like a job indicator script, the AI couldn't track that person since the name it would try to track would literally be "John Doe [Assisstant]" instead of "John Doe". This should fix that.

Tested on a test server and the AI can track people just fine now. This doesn't affect voice changers in any way that I can find. Looking at it now I'm not entirely sure if html_encode() is necessary, but it's late and I need sleep. I'll fix it tomorrow if html_encode() is needed, but it should be fine.


danhibiki:
good shit

Nokington:
:+1:

TDcoolguy300:
Not sure what this might do, i am not familiar with NTSL

PR #1664, opened 2016-01-17
Buffs moonlight sword

danhibiki:
Now can shoot as long as it has one charge. Shot type is dependent on how many charges you have, with more charge = better shot. All charge is expended on any shot, and melee force also scales with charges as well, going from a low of 20 with no charges and 45 with charges.


Nokington:
sure

PR #1666, opened 2016-01-17
Defines a name for proc queue

danhibiki:
Doesn't have one now so it shows up as a nameless statclick, this is just so things look pretty. Last PR wasn't right


danhibiki:
Not even needed, going to close this and make a new PR to remove the one check we have for procqueu

PR #1667, opened 2016-01-17
Removes proc queue from subsytems

danhibiki:
Totally unused subsystem that was only called in crew monitoring, referred to https://github.com/tgstation/-tg-station/blob/master/code/game/machinery/computer/crew.dm to see how to do without it


Nokington:
sure i guess

PR #1740, opened 2016-01-25
Readds Sheet Solidification

ArcLumin:
Revert removed it, readding it.


danhibiki:
Just incase you weren't aware, we are currently not merging any PRs to make sure we've properly identified our source of crashing. You can keep making PRs, just don't be alarmed if there is a lack of immediate response

ArcLumin:
Makes sense. Didn't know about that. I've had issues with slack so I haven't been using it. I know for sure these things won't cause issues, but of course, it's a lot easier to block all. I'll just leave these here for now then.

Nokington:
Once the freeze is over, :+1:

PR #1746, opened 2016-01-26
Adds the map changes we used to have

Pyko1:
The lagging is almost 99.9% sure not a mapping issue.
This re-adds all the fixes and content that had been added to map, excluding mime statues, as they were invisible or something.

Also includes roundstart AT fix to z2.


Pyko1:
Made the nukeops blastco access a bit more bomb/emag proof.

capture

FuryMcFlurry:
that is hilarious.

Pyko1:
I think you could still get past it with 16 c4's.

FuryMcFlurry:
make it a teleporter entrance then. activate it and you can teleport to the blastco room.
Or make c4 unusable in the Z level.

danhibiki:
We can just solve the blast door thing thru code, defining an emag_act that does nothing I'm pretty sure. The code hiatus isn't going to be permanent, just hang on to your butts

Pyko1:
Well the button has been made unemaggable in this PR, only explosions can be an issue.

PR #1750, opened 2016-01-27
Adds LLA templates again

TDcoolguy300:
Also brings in all active templates that were given to me at the time.


Nokington:
Once the freeze is over, :+1:

Lawligagger:
Nice

PR #1761, opened 2016-01-28
wrong damtypes for switching damage types

sp1d3rx:


nalarmg:
Not entirely sure why each file changed has it's own PR. If possible try to avoid that please! When working on a project like this, submitting them under one PR is preferable.

sp1d3rx:
yeah 1764 has all in one, github was being weird. I did finally get it working though.

PR #1764, opened 2016-01-28
Mega crash fix based on logs 

sp1d3rx:
travis please work, thanks.


sp1d3rx:
can someone please kick travis again? the code is fine, it's travis that is not doing his job.

sp1d3rx:
this also fixes that addlog null.x error - if there's no turf found at the target location it will say (invalid target location) instead of logging nothing.

sp1d3rx:
There may be 32 files changed but mainly it's null checks and using constants instead of hard coded values.

PR #1773, opened 2016-01-30
Fixes shadowlings dropping organs and implants on hatch

Carbonhell:
Pretty urgent bugfix because this can be exploited if you know how.


Nokington:
yeah sure :+1:

PR #1782, opened 2016-01-31
Map fix

Pyko1:
Fixed the improper access on hidden airlock next to commie spy and on the door leading to chapel back room.


Pyko1:
Fixed sec holding cell disposal going through holodeck, as well law office disposals being connected to it.

Pyko1:
Fixed the sec escape pod having space tiles under some walls during transit.

Pyko1:
Fixed Roundstart AT at Commie spy hidden airlock.

These are all just fixes, so would be nice if these could be merged soon, as I was planning to start mapping stuff on sec and it will most likely conflict with some of the fixes due to being at same X-coordinates.

Pyko1:
These are included in https://github.com/HippieStationCode/HippieStation13/pull/1787

PR #1790, opened 2016-02-02
Arrivals Revert

AliiTea:
https://trello.com/c/M9jtGfL2/686-revert-arrivals-to-the-original
Reverted Arrivals to it's Original form.

Before
before

After
after


Pyko1:
Maybe add those extra colored clothes closets somewhere around tho?
Even tho, I think most of the issues with "new" arrivals were more or less resolved.
Also I recall there was some bad camera coverage at hallway next to recreation, that I fixed on live,  so would be neat if you could toss a camera around there.

Pyko1:
Also, might want to connect those support lattices to catwalk.

Kokojos:
Yes :+1:

Just make sure there is no errors with atmos and areas and such.

Nokington:
I say we wait a bit and see how new arrivals is now that Pyko made it better.

ghost:
:-1: I like the current arrivals

Spacedong:
:-1: those catwalks look like ass, add some lattices and connect them to station

TrashFiend:
:+1: Current arrivals is and has always been garbage, old arrivals is far superior. Catwalks are a bit far from the station, though.

Spacedong:
arrivals is currently fine and workable unlike the brig currently, I'd focus on redesigning the brig and adjusting the brig shuttle to actually work (with the prisoner section IN THE BACK and the blue area in front with all the controls).

TrashFiend:
Current arrivals is still large for no reason, though. The fancy new Security Office and conveyor stuff gets extremely little use and the Invasive Searching Area gets even less. On something like Bay I could see those changes adding to anything but there really isn't any reason to stay there when Dickbag McGee will just break the window and rush through if you try to use it properly. The pods being further away from everything isn't exactly fun either.

Brig is totally worse, though.

Spacedong:
Yeah Arrivals was originally re-designed to look more RP-oriented on a server with little to no RP. I complained about this before it was merged. The problem is that theres that one little area to the right above shuttle that only has one cig vending machine and a stool. The pods are so far away that its ridiculous. The office with the Comms Console also can be broken in through 3 directions.

Spacedong:
Another thing I don't understand is that the Comms Console Office is supposed to be for ID changes yet doesn't have shutters or even a counter to put ID's on.

blobbernaut:
this just makes less station. I'm bored of the old arrivals look. It's been the same for like 3 years +

The current arrivals is just refreshing. If you want to you can just remove the rubber flaps and conveyor belts, but I like the layout as is. 

[thumb down emoji I dunno how to make]

Nokington:
Even if we merged this it has conflicts.

AliiTea:
I can see that Nokington, in 2 hours I can updoot it.

AliiTea:
@Nokington going to open a new one without conflicts

PR #1800, opened 2016-02-04
Adds spamcheck for fedora tipping

danhibiki:


TDcoolguy300:
Shure

Nokington:
:+1:

code/modules/clothing/head/misc.dm
```diff
@@ -180,19 +180,25 @@
...
 	var/mob/living/carbon/human/H = user
-	if(user.gender == MALE)
-		user.visible_message("<span class='italics'>[user] tips the [src]! It looks like they're trying to be nice to girls.</span>")
-		user.say("M'lady.")
-		sleep(10)
-		H.facial_hair_style = "Neckbeard"
-		H.adjustBrainLoss(10)
+	if(H.spam_flag)
+		return
 	else
```
nalarmg, line 185: Not an issue, but just a note: this else isn't actually needed, since it'll just continue after this point anyway with or without it.

PR #1810, opened 2016-02-05
Run Map Merge Settings

Pyko1:
Changed the run map merge target location from "_maps" to "_maps/map_files". 
Generally only those maps should need mapmerge, and the prepare maps only makes backup of the maps in map_files folder anyways, so map merging wont work for others.
Also the list is starting to get ridiculously long with templates.


Nokington:
:+1:

danhibiki:
:+1:

Issue #1696, opened 2016-01-22
shards still embed in golem feet


Spacedong:
just give em chitin shoes

Cherkir:
lmao truuu

Issue #1055, opened 2015-10-27
Golems appear to show the "No Oxygen" symbol on their HUD when in a room without oxygen, but not while in space


TrashFiend:
Not sure if this was a one time thing or it even matters, but as a golem the hud element saying that there isn't enough oxygen only shows up in rooms, not in space. Not a big issue in the slightest, but still a bit weird.

Pyko1:
Humans seem to be missing the no oxygen symbol in the space as well.

danhibiki:
Whats the status on this?

Pyko1:
Should be fixed.

PR #1841, opened 2016-02-08
Golem fix

Pyko1:
added specflags = list(NOBREATH,VIRUSIMMUNE,NOBLOOD,PIERCEIMMUNE,COLDRES, HEATRES, NOFIRE, NOGUNS) to adamantine golems.
Some of these things used to be provided by clothes, which they don't apparently anymore have.
This means they take reduced damage from heat/cold (including lasers) and cant be set on fire, they wont bleed or catch a disease and neither can syringes be used on them, and they wont take damage for walking on dices or shards, or hitting things with shards. But as a downside, they are now too bulky to use guns.

Removed meat = null from meeseeks (as it also has meat = /obj/item/weapon/reagent_containers/food/snacks/meat/slab/human/mutant/meeseeks)

Made the glass shards check for if(!(PIERCEIMMUNE in H.dna.species.specflags)) when crossed.

Fixes: https://github.com/HippieStationCode/HippieStation13/issues/1696 and https://github.com/HippieStationCode/HippieStation13/issues/1055

### DIFF
--- a/code/game/objects/items/stacks/sheets/glass.dm
+++ b/code/game/objects/items/stacks/sheets/glass.dm
@@ -329,21 +329,22 @@
 		playsound(loc, 'sound/effects/glass_step.ogg', 50, 1)
 		if(ishuman(AM))
 			var/mob/living/carbon/human/H = AM
-			var/check = H.lying ? H.w_uniform : H.shoes
-			if(!check)
-				var/obj/item/organ/limb/O = H.get_organ(pick("l_leg", "r_leg"))
-				H.apply_damage(5, BRUTE, O)
-				if(!H.lying) H.Weaken(3)
-				if(prob(embed_chance))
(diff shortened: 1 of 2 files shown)


Pyko1:
I guess the flags can be considered as features as well even the NOBREATH and PIERCEIMMUNE are used to fix issues. Also, due to their bulky and stoney form. Should there be other flags such as COLDRES, HEATRES, RADIMMUNE, NOFIRE and/or  NOGUNS?

Cherkir:
Can't check the code right now, but :+1:. Golems should be made faster as well IMO.

Kokojos:
Damnit I already have the second fix made I just suck enough with git not to know how to PR from a branch only :P

Nokington:
:+1:

danhibiki:
Yes

PR #1847, opened 2016-02-08
Welder fire alert fix

Jaygriff:
Fixes admin log spam due to welder
fixes #1832

It will not alert admins 3 times of welders lit in areas with .01% plasma, then wait 1 minute, then alert admins again. This time can be adjusted.

Issue #1832, opened 2016-02-07
Welding logs happen when a borg is being repaired


Kokojos:
![bug](http://s9.postimg.org/oogw0eygf/bugreport.png)
It's the guy in the top left.

Kokojos:
and also apparently everytime it's used

Pyko1:
IT should be changed to be releated to hotspot/new() or something.

Kokojos:
AHHHH IT'S BEING USED EVERYTIME SOMETHING HAPPENS IN AN AREA EVEN IF THERE IS 0.0000001 plasma.

TDcoolguy300:
Carbon's fix caused shit.

PR #1853, opened 2016-02-09
Welder Logging Improvements

Jaygriff:
Ported conflicting fixes from #1847, 
expanded fix for #1832


Nokington:
:+1:

Carbonhell:
Not sure,instead of a spamcheck wouldn't edit this check "if((air_contents.toxins > 1)" from 1 to the amount needed to create a fire?Wouldn't that kinda solve everything

Jaygriff:
It still would continue to spam the admins when its in a room with plasma, 4 warnings is enough for a admin to realize and do something about it, if they were going to.

PR #1856, opened 2016-02-10
Nerfs borg stun arm

blobbernaut:
It now costs 300 power instead of 30 power, which is basically free.

For comparison, the sec borg's stun baton costs 1000 power. Why a tool specifically designed to incapacitate humans with nonlethal electric shock costs 30 TIMES MORE POWER than the stun arm, I don't know.

Quite frankly the stun arm makes no sense. Realistically it'd just char the person to a crisp before it shocks them.

### DIFF
--- a/code/game/objects/items/robot/robot_items.dm
+++ b/code/game/objects/items/robot/robot_items.dm
@@ -10,7 +10,7 @@
 	icon_state = "elecarm"
 
 /obj/item/borg/stun/attack(mob/living/M, mob/living/silicon/robot/user)
-	if(!user.cell.use(30)) return
+	if(!user.cell.use(300)) return
 
 	M.Weaken(5)
 	M.apply_effect(STUTTER, 5)


TDcoolguy300:
I recommend adding a sound too

trifornt-tinfornt:
It's a change that'd be welcome to be honest, I'm not the largest fan of nerfing borgs other than secborg (rip mediborg) but honestly stunarm can be used to murderbone waayyy too easily.

Nokington:
:+1:

blobbernaut:
Hmm. I'll increase the power cost of the laser gun too, and buff borg spray. Maybe change this from 300 to 200 power cost. I don't know what caused polyacid to be utter shit, would creating a new chemical with much higher acidpower do the trick?

blobbernaut:
The only reason I want to nerf stun arm is because it makes emagged engi borgs good at literally everything in the game, while also being unstunnable except at melee range, or with the ion rifle, of which there is only one on the station.

You play borg a lot, trifornt. What is your perspective on all the borgs, and do you have suggestions for the others?

trifornt-tinfornt:
The power of the laser gun wasn't the problem (it already steamrolls through energy, not even exaggerating a little bit if you're in a fight with like, three people you'll be out of energy pretty fast especially if your aim sucks) the problem comes from the ranged power it has combined with the on-foot speed of a 100% human, and how much better it is alone. Yes, the Engiborg's versatility when now having a stun is FANTASTIC, but the stun puts him at risk of flashes already and if you're chasing a guy who is at 100% then you're not gonna get him. I'd say a 200 Power Cost would be more fair in the long run.

Yes I do have ideas (complaints) about the other borgs, but honestly this isn't the place to discuss them. Guess I'll just make another topic in the suggestions.

Carbonhell:
Imo 300 is good, 30 is way way too small, you're not supposed to stunspam with it

danhibiki:
300 is a good start, lets see how it goes.

PR #1860, opened 2016-02-11
Adds in colored reagents + Bleach, more grinding

ArcLumin:
-Adds in colored reagents with specific colors made via colorful reagent + same type crayon powder
-Adds in Bleach, which works like space cleaner, but is toxic on ingest/inject, and can remove colorful reagent from mobs, tiles, you name it
-Adds in coin grinding, so you can turn them into that type of reagent
-Adds in grinding the clown's crayon for some colorful reagent

Note: I tested making an invisible reagent with the mime's crayon, but it made all mobs, items, you name it that it touched invisible, and made floor tiles completely black. A little too overpowered. So, it was removed.


ArcLumin:
Thanks travis, now I gotta work on this shit.

PR #1863, opened 2016-02-11
Adds in colored reagents + Bleach, more grinding

ArcLumin:
-Adds in colored reagents. Now you can force your colorful reagents to be a specific color! Paint the station red! Simply add crayon powder to colorful reagent
-Adds in Bleach. Not only is it painful to drink, but it'll remove stains like Colorful Reagents.
-Adds in more grinding to the chem grinder, such as coins and the clown's crayon


ArcLumin:
Goddamnit travis

Issue #1859, opened 2016-02-11
Empty tasers display 1 shot left on the player's sprite.


Spacedong:
http://vgy.me/cyKIu6.png

yacobm8:
Working on it now and it also seems to only bug out when you have 2 shots in the taser.

When charged to 2 shots it sometimes only fires 1 shot. (Possibly if you only just charge it to 2 shots instead of towards the end of the charge of 2 shots).

It leaves the taser empty in your hand but, on your character in game it appears to have 1 shot left.
Here's a video of the bug
https://www.youtube.com/watch?v=liWhm5ouSv0&feature=youtube_gdata

yacobm8:
This bug also affects all other energy weapons in that when they are emptied of ammo they wont flash to show they are out of ammo I should have a pull request up tomorrow for this

Kokojos:
I think it's mostly their powercell not being exactly 0, but between 1 and the energy neeeded to shoot (100)

yacobm8:
Hey Kokojos, I've fixed the problem it was a problem with overlays. I'm unable to get github to work though on my desktop. I've tried to upload my new files to my forked version but I don't know why it's not working. Could i give you the files I've changed and you take a look?

Kokojos:
If you give me the correct files plus a changelog, I could PR it for you by tomorrow.

Otherwise sure I can take a look but Im no expert, especially in github.

yacobm8:
Okay, no problem I'll write a changelog now. How do you want me to send you the files?

Kokojos:
My email is in my profile.

yacobm8:
So I just got emailed back from github support who told me how to get around the issue I was having (great timing). I've uploaded the change to mine now and will write a change log and do the PR :) Thanks for offering to help me out

yacobm8:
I've proposed a fix for this in pull request #1878

Carbonhell:
Tasers are fuckingly weird, a guy's taser's cell got its charge reduced to 50 causing it to show up as with 1 charge left but not being able to shoot,tasers are weird

Carbonhell:
@Spacedong try this

Gandalf2k15:
This bug was apparently fixed. http://i.imgur.com/wxA8a25.png

Kierany9:
This bug was fixed

PR #1878, opened 2016-02-15
issue #1859 Weapons now show correct sprite/overlays respective to how much ammo they have left.

yacobm8:
I've changed the way the old system works a little because it's the only way I could think to fix this issue. I'm not experienced with SS13 and would feel better if somebody else took some time to look at the issue first before committing my fix. Maybe there is an easier way to fix the problem but here is my solution. 

If I’m right I think the problem was inside energy.dm in the update_icon function the main line causing a problem is 
`var/ratio = Ceiling((power_supply.charge / power_supply.maxcharge) * 4)`
If you charge your taser to 300 it will show you having 2 shots left even though you only should have 1 shot because for stun ammo it's 200 per shot. That line of code reads... 
`var/ratio = Ceiling((300 / 1000) * 4)`
The answer is 2 after you round up 1.2.

This ratio variable is used to determine how many shots to add to the overlays and sprites on a weapon which is why it shows you having 2 shots instead of 1 left. So when you fire your last shot its not showing the correct sprite for an empty weapon because, now your charge will be 100 making the ratio variable equate to 1 (rounding up the number 0.4).

So the current system is always at risk of rounding up the number to give a false sprite. My way of doing it doesn't work out a ratio and then round the number to determine how many overlays to add. It instead takes the number of shots left in the cell and then adds the correct number of overlays based on the weapon being used. 

This means that I've had to add the variable overlay_ammo to energy guns. This variable is the number of ammo a weapon can hold on its sprite. So for example tasers have 4 while the stun revolver only has 1 visible overlay for ammo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 *.txt
 *.int
 *.db
+code/game/gamemodes/abduction/abduction_gear.dm
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,3 @@
 *.txt
 *.int
 *.db
-code/game/gamemodes/abduction/abduction_gear.dm
--- a/code/modules/awaymissions/mission_code/stationCollision.dm
+++ b/code/modules/awaymissions/mission_code/stationCollision.dm
@@ -66,6 +66,7 @@ obj/item/weapon/gun/energy/laser/retro/sc_retro
 	desc = "An older model of the basic lasergun, no longer used by Nanotrasen's security or military forces."
 //	projectile_type = "/obj/item/projectile/practice"
 	clumsy_check = 0 //No sense in having a harmless gun blow up in the clowns face
+	overlay_ammo = 4 //number of sprite cartridges visible. Used in energy.dm to switch sprites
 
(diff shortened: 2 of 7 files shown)


Carbonhell:
Couldn't you just swap Ceiling with Floor() which takes the lower value?

yacobm8:
I did try that straight away and for a reason I can't now remember it didn't work for me. p.s the floor function is round in byond

nalarmg:
It would honestly be better to fix the given ratio, or maybe think of an alternate. This is a not nice fix. Not only are you adding a new variable to every gun that needs its sprite updated when discharging, but you also have maintenance overhead where you need to update the variable anytime you want the amount of shots to change.

Coders, thoughts?

yacobm8:
Yeah I totally agree nalarmg I think then that my explanation of the problem is useful for who wants to make a fix but this pull request shouldn't be merged.

Nebyab:
oh
is that a request to close it or

yacobm8:
Yeah you can close it. I'm a bit lasy to do it again

PR #1897, opened 2016-02-20
Reverts part of the noose update.

danhibiki:
I think its the animate here thats causing our sudden issues and lockups.


Nokington:
aeyh eusr

PR #1914, opened 2016-02-22
2 new adminsverbs : Clean Air and Fix Breach

Kokojos:
Clean air : Cleans all the air in an area. Allows to instantly stop plasma fires. requires usually 2 uses because the canisters are emptying. Can also clean N20 and superheated air.

*edit : Stolen mostly from LLA, of course, then modified by yours truly.

Fix breach : Replace all space in the desired area to be replaced with plating.

firesssst


Spacedong:
rip lame pingas

Nokington:
do it

PR #1919, opened 2016-02-25
This adds logs.

Kokojos:
Fire spread logs.
Gun intended targets
better ''tabled'' verb instead of pushed (seriously TG)

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -201,7 +201,7 @@
 
 /obj/item/weapon/gun/proc/process_fire(atom/target as mob|obj|turf, mob/living/user as mob|obj, message = 1, params, zone_override)
 	add_fingerprint(user)
-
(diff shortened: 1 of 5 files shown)


Nokington:
sure

code/modules/projectiles/gun.dm
```diff
@@ -201,7 +201,7 @@
 
 /obj/item/weapon/gun/proc/process_fire(atom/target as mob|obj|turf, mob/living/user as mob|obj, message = 1, params, zone_override)
 	add_fingerprint(user)
-
+	add_logs(usr, target, "shot at", admin=0)
```
Carbonhell: um no, why are you using usr instead of user which is in arguments,plus what the hell is that admin parameter

Kokojos: Il be honest, I litteraly copy pasted my earlier changes.

Thank you for spotting that.

PR #1924, opened 2016-02-26
Adds ProxyKick

Carbonhell:
Terms of Service
By using this service, you agree to:
Not sell this service or information generated from this service, directly or indirectly, without explicit consent.
Not use / reuse information generated from this service, directly or indirectly, without giving credit to the source (this website).
Not exceed the query limits if you're a free user.
Not look up random IPs / incremental IP lookups. The database changes very often so information becomes stale very fast. It just causes a higher server load for no reason.
The Terms of Service may change at any given time, without prior notice.

All credits to http://getipintel.net/ which is used in the code for the Proxy checker.


Carbonhell:
> hog updates snuck in
> fuck
> gotta wait for hog pr to be merged first

PR #1931, opened 2016-02-28
Makes gang tags lower layer if on a floor.

### LINKED ISSUE #1886: Graffiti sprayed on the floor appears over lockers
http://i.imgur.com/7Aedz5K.png

Pretty minor, but still

Kokojos:
gangtool
Attemps to fix #1886

Obviously this gives a big advantage to gang members. An open lockers almost covers the entire thing, but at least it's not ugly as fuck.

headmins shall decide.

### DIFF
--- a/code/game/objects/effects/decals/crayon.dm
+++ b/code/game/objects/effects/decals/crayon.dm
@@ -36,6 +36,9 @@
 	var/datum/gang/gang
 
 /obj/effect/decal/cleanable/crayon/gang/New(location, var/datum/gang/G, var/e_name = "gang tag", var/rotation = 0)
+	var/target = get_turf(src)
+	if(istype(target, /turf/simulated/floor))
+		layer = 2.9
 	if(!type || !G)
 		qdel(src)
 


Pyko1:
Other solution could be restricting them only on walls.

Carbonhell:
Why checking if the gang spray is being made on a floor?Can't you set default layer to 2.9?

Pyko1:
I'm starting to think it might be better to restrict tags on walls only.

2.9 Is under doors, computers and pretty much all machinery. 
So it will be impossible to see tags that are sprayed under computers/venders and such.

Carbonhell:
why not raise it more then?

Pyko1:
then its back above lockers, lockers are 3.0 if i remember right.

Kokojos:
Yeah, most ''things'' are above 3.0, I'd have to check for the floor to have no objects-  but then you could move one right over.

Nebyab:
I mean alright?

Pyko1:
Should have at least a check for anchored machinery, at least then you would have to build it over the tag, instead of just spraying under it.

Nokington:
It would be better to just make it so they can only be placed on walls

PR #1933, opened 2016-02-28
carbon is a cuck

Pyko1:
ask carbon, hes the cuck here.


Carbonhell:
Aka fixes the "You're brainwashed,you're no longer a follower!" message when borging an antag, mergebeg

PR #1935, opened 2016-02-28
Fixes last chemical implant bug

Sarcalogo:
Now they can no longer be self activated under any circumstances, only through the console.

### DIFF
--- a/code/game/objects/items/weapons/implants/implant_chem.dm
+++ b/code/game/objects/items/weapons/implants/implant_chem.dm
@@ -32,12 +32,6 @@
 	tracked_implants -= src
 
 
-
-
-/obj/item/weapon/implant/chem/trigger(emote, mob/source)
-	if(emote == "deathgasp")
-		activate(reagents.total_volume)
-
 /obj/item/weapon/implant/chem/activate(cause)
 	if(!cause || !imp_in)	return 0
 	var/mob/living/carbon/R = imp_in


Chnkr:
How is it a bug when it's a feature?

Sarcalogo:
It's supposed to activate on death. It actually activates when you deathgasp. People are going around with the implants installed and just copy pasting deathgasp in their chat bar to activate it. It's an oversight, but still. There is no real purpose of it activating when you die. If you're cloned you're in a new body, so the implant is gone regardless.

Pyko1:
I think its there mainly for changelings.

Sarcalogo:
The issue is still present, power gamers going around with implants that heal them of a shit ton of damage, remove stuns, revive from death, and can activate at any time. Nobody uses them for their actual purpose. This is the last thing that needs to be fixed to stop it.

Carbonhell:
Code's good, but i wanna be sure of a thing:Does the action button pop up, not functional?Or doesn't it pop up at all?the latter should be the effect wanted,but the code is kinda...confused

Sarcalogo:
I removed the action button in a previous PR. It doesn't pop up at all.

 This removes the ability to simply fake deathgasp and activate the implant remotely. If you know of a way so that it activates on death, and not deathgasp, I'll change it so it works that way.

Pyko1:
Basically this at -100 health: https://github.com/HippieStationCode/HippieStation13/blob/7e8e4e70be2bf67c1be5800ef7fd0f66ff50efd6/code/modules/assembly/health.dm
And then just add in some feature for changeling regen death thinggy,

Carbonhell:
Just keep in mind that chemicals don't metabolize on corpses.

Pyko1:
Does not really matter, both of the cases used to trigger it for what ever reason.

So I assume its to get rid of it if you die and get revived with what ever.

Nebyab:
I was under the impression that it was supposed to activate upon being crit, as that's how it used to operate.

Though the idea of playing dead somehow making a chemical implant go off is hilarious, it makes no sense and perverts its intended purpose.

Pyko1:
the *deathgasp suggests it is supposed to activate on death like bomb implants.

Nebyab:
That might've been a change that accompanied defibs then

Nebyab:
@Nokington oi dickface

Nokington:
WHAT

Spacedong:
MERGE THIS WHAT ARE YOU WAITING FOR 

REMOVE POWERGAMING IMPLANTS ONCE AND FOR ALL

DENTAL IMPLANTS, YOU'RE NEXT

Nokington:
:+1:

Nokington:
I'm actually not sure about this because you should always be able to *deathgasp to trigger explosive implants. But if they just triggered on crit that would be good too

Nebyab:
Unless I'm horribly horrible mistaken, this only touches the Chem implants.

Nokington:
As long as it only touches chem implants then :+1:

Spacedong:
this is literally a bug fix and not a feature, if anything merging this will FIX balance, along with removing dental implants plz

Nokington:
@Nebyab Are you for this or against this

Nebyab:
I was the person who originally said that this is a bug. I was also the person that asked for this to happen. I thought my approval was implicit.

PR #1936, opened 2016-02-28
Adds ProxyKick system

Carbonhell:
DO NOT MERGE.
Proxychecker uses this, all credits goes to this wonderful tool's creators and Bartels for suggesting it!
http://getipintel.net/
Also needs config editing.

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -191,6 +191,11 @@
 	var/tries = 10
 	var/directory = null
 
+	//proxykick
+	var/proxykick = 0 // disabled by default
+	var/proxykickemail = ""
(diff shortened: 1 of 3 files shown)


code/controllers/configuration.dm
```diff
@@ -387,6 +392,12 @@
 					protected_config.autoadmin = 1
 					if(value)
 						protected_config.autoadmin_rank = ckeyEx(value)
+				if("proxykick")
+					proxykick = 1
+				if("proxykickemail")
+					proxykickemail = value
+				if("proxykicklimit")
+					proxykicklimit = value
```
NullNeska, line 400: I'm almost certain this should be text2num(value)

PR #1939, opened 2016-03-01
Refractors exosuit fabricators

Carbonhell:
Just a code fix so it uses the materials_container datum instead of some snowflake stuff, no feature here


Nebyab:
uh what
sure alright

PR #1959, opened 2016-03-05
Adds debug message for proxycheck

Carbonhell:
mergebeg


nalarmg:
This shouldn't show up to players, which there's no check for.

Carbonhell:
I'll only use on my own client since it's just for checking, it'll be removed after i use it

Carbonhell:
Plus the proxychecker is disabled and can only be called by Call proc verb by an admin

nalarmg:
I'd rather there be a holder check as a fail safe.

Carbonhell:
replaced src<< with message_admins

PR #1960, opened 2016-03-05
Fixes bad indentation

Carbonhell:
Fixes bad indentation in deathsting code, which was causing warnings. Why doesn't trello check if DM gives warnings?


Pyko1:
code seems fine, and this is a fix.

Merging.

PR #1692, opened 2016-01-21
Monkeys not being able to be humanized

Carbonhell:
I fucked up a check ayy
Fixes #1690

PR #1962, opened 2016-03-06
Monkey fix

Pyko1:
Re adds carbons fix for monkeys not turning to humans.
https://github.com/HippieStationCode/HippieStation13/pull/1692/files


Pyko1:
Needs code review mainly if there was some actual reason why the fix was reverted.

Carbonhell:
Yea it's ok, i accidentally put it cause nearly all stings need to check for ..(), but this is an exception,code's ok

PR #1976, opened 2016-03-07
Buffs the RD's Hardsuit

AliiTea:
After the reactive armor was removed RD was nerfed.. so this should help to make RD not nerfed.


Nebyab:
I mean, I don't feel as though its needed strictly speaking. To claim that the nerf of the telearmour is an RD nerf is a bit of a misnomer, as its really more "making things like they used to be: i.e not terrible and nonsensically balanced". The heads are inherently unequal in gear, we don't really need to buff the RD's hardsuit, well at least from my perspective.

Nokington:
Yeah sure

Carbonhell:
No. RD. doesn't. need. an. armor.
If you really want to improve RD somehow, find a way to make him get minerals in a pretty hard but still possible and less shit than going mine yourself way...perhaps by using the gooncode disk,unsure

Nokington:
this doesnt matter to me either way. I dont play rd and people that do are nerds

GunHog:
I am not a Hippie, but I think it is more _Science_ appropriate if R&D could produce upgrades to hardsuits. Maybe even the CE or HoS would drop by the lab for some kit as well. Just some input because Nexie likes poking me :P

Examples from /tg/: Red slime potion: Any clothing item with slowdown has it removed + vehicles go FASTA.
Slime Chill potion: Fireproof any clothing item.
Hardsuit Jetpack upgrade: Can be bought by miners to give themselves a hardsuit jetpack.

Zadeon:
Lets just turn hardsuits into powersuits with upgrades that R&D can build. The hardsuits could have a grid based module system where upgrades take up so many squares on the grid.

GunHog:
I believe Bay had a thingie like that.

Carbonhell:
Paradise,rather,and yeah it'd be nice if the RD armor worked on modules rather than just buffing it. So -1

Nebyab:
Giving the RD better armour, in a role perfectly suited for something wacky and neat, is lame. Its also not needed. The RD got along fine for years without armour, we should take this as a chance to do something really neat, not just buff a hardsuit.

Carbonhell:
PR rejected by an headmin and several people, closing it

PR #1982, opened 2016-03-09
Conveyor switch

Pyko1:
Made conveyor switch update the conveyor belts soon after the map is loaded.

Meaning that you just have to mapedit the switch on/off, and it will pass it on to linked belts soon after the map is loaded, instead of having to edit the switch and all the belts.

Technically a fix to disposals conveyors starting off even tho the switch is on, but also a mapping feature.

### DIFF
--- a/code/modules/recycling/conveyor2.dm
+++ b/code/modules/recycling/conveyor2.dm
@@ -207,6 +207,8 @@
 		for(var/obj/machinery/conveyor/C in machines)
 			if(C.id == id)
 				conveyors += C
+		spawn(5)
+			operated = 1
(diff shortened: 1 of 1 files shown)


Nokington:
Sure

Pyko1:
Basically the operated thing is used to transmit on/off and the move dir to the belts.
The spawn(5) is mostlikely unneeded, but i had some just incase reason when i made it.
Even tho it affects all (new) conveyor switches, only case where it actually does anything is when a switch is spawned with edited vars.

Nebyab:
sure

PR #1994, opened 2016-03-11
Ports from #12431 to #12465

Carbonhell:
Most important port is the overlays update which allows for some fun stuff and removes a lot of copypasta code!
Also removes plant race which was completely unusued (keep in mind that plant and pod are two different things!)


Nokington:
wew lad we have reactions now on github

Carbonhell:
Remember,this is mainly bugfixes/refactoring, the maximum you can see is an item being added which is not accessible ingame other than asking an admin to spawn it(Laser eyes injector)!

TDcoolguy300:
Looked through it, codes fine, it's a refactor and bugfix so i am going to go and merge

PR #2006, opened 2016-03-13
Fixed changeling hive absorption

JohnGinnane:
Now checking if the selected DNA name is already in our stored profiles.

### DIFF
--- a/code/game/gamemodes/changeling/changeling.dm
+++ b/code/game/gamemodes/changeling/changeling.dm
@@ -366,11 +366,12 @@ var/list/slot2type = list("head" = /obj/item/clothing/head/changeling, "wear_mas
 
 	var/datum/changelingprofile/prof = new()
 
-	H.dna.real_name = H.real_name //Set this again, just to be sure that it's properly set.
+	H.name = H.dna.real_name //Set this again, just to be sure that it's properly set.
 	var/datum/dna/new_dna = new H.dna.type
 	H.dna.copy_dna(new_dna)
 	prof.dna = new_dna
-	prof.name = H.real_name
+	prof.dna.real_name = H.name
+	prof.name = H.name
 	prof.protected = protect
 
 	prof.underwear = H.underwear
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/changeling/changeling.dm
```diff
@@ -366,11 +366,12 @@ var/list/slot2type = list("head" = /obj/item/clothing/head/changeling, "wear_mas
 
 	var/datum/changelingprofile/prof = new()
 
-	H.dna.real_name = H.real_name //Set this again, just to be sure that it's properly set.
```
JohnGinnane, line 369: This seems to be causing an "undefined variable" error which is why you couldn't absorb through hivemind in the first place. I tested absorbing normally and through hivemind and both are working

code/game/gamemodes/changeling/powers/hivemind.dm
```diff
@@ -81,11 +82,18 @@ var/list/datum/dna/hivemind_bank = list()
 
 	var/S = input("Select a DNA absorb from the air: ", "Absorb DNA", null) as null|anything in names
 	if(!S)	return
+
 	var/datum/changelingprofile/chosen_prof = names[S]
 	if(!chosen_prof)
 		return
 
+	for(var/datum/changelingprofile/stored_prof in changeling.stored_profiles)
```
JohnGinnane, line 90: Here I'm checking if the selected person isn't already in your stored profiles using their name, as when you channel a DNA it creates a copy which isn't equal to the one in your stored profiles. This meant you could potentially "absorb" the same person and get free changeling points!

code/game/gamemodes/changeling/powers/hivemind.dm
```diff
@@ -71,9 +71,10 @@ var/list/datum/dna/hivemind_bank = list()
 /obj/effect/proc_holder/changeling/hivemind_download/sting_action(mob/user)
 	var/datum/changeling/changeling = user.mind.changeling
 	var/list/names = list()
-	for(var/datum/changelingprofile/prof in hivemind_bank)
-		if(!(prof in changeling.stored_profiles))
-			names[prof.name] = prof
+	for(var/datum/changelingprofile/hivemind_prof in hivemind_bank)
+		for(var/datum/changelingprofile/stored_prof in changeling.stored_profiles)
+			if(stored_prof.name != hivemind_prof.name)
+				names[hivemind_prof.name] = hivemind_prof
```
JohnGinnane, line 77: This is kinda dumb I think because you'll still populate names. Even if you selected someone you've already absorbed it's handled down below

Issue #1343, opened 2015-12-01
Tracking Beacon microphones


ghost:
they do not work,but the speakers do

Spacedong:
lmao who the fukk wud care bout this shit u autist

TDcoolguy300:
Not a bug, intended in the code.

carlarc:
remove the microphone then, it looks bad.
also how did this get ignored for so fucking long?

Carbonhell:
Fixed in #2013

PR #2013, opened 2016-03-17
Beacon Microphone Fix Balances #1343

Zadeon:
This is just a balances but fixes the microphones on tracking beacons so that they work. 
I was looking for issues I could maybe fix and saw this one so here it is.

### DIFF
--- a/code/game/objects/items/devices/radio/beacon.dm
+++ b/code/game/objects/items/devices/radio/beacon.dm
@@ -6,13 +6,6 @@
 	var/code = "electronic"
 	origin_tech = "bluespace=1"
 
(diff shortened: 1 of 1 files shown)


TDcoolguy300:
As above, balances #1343

Nebyab:
yeah sure

Carbonhell:
To be fair,this is a sloppy fix. Tracking beacons shouldn't even be radio childs at all, making them work like radios sounds just...weird, no clue.It'd be good if someone reworked them to be their own item, but code's fine anyway i guess.

Carbonhell:
So yeah,i'd prefer if you actually made it its own item instead of being a radio child. But if you want this to be merged, say so, code's fine

Nokington:
:+1:

PR #2015, opened 2016-03-17
Fixes Templates

TDcoolguy300:
Two map files were causing the issues with very little known cause, removed the affecting maps and added a Generation button to the template panel


TDcoolguy300:
Also, http://puu.sh/nJkdd/c9a089979e.png

Nebyab:
someone codecheck so this can be merged

Pyko1:
Might had been better to just move those out of the random space generation folder until the problem is fixed.

Mapwise those should be ok

Carbonhell:
code is ok

Carbonhell:
It's a bugfix,merging this.

PR #2016, opened 2016-03-17
Fix for Multiple Messages when Farting on Somebody

Zadeon:
This just fixes both messages of you farting and you farting on somebody displaying when you do fart on somebody. Now it'll only show you farted on somebody. 

Also cleans up some old code being used.


Carbonhell:
codewise it's ok,not sure how many people consider this as a bug through

Zadeon:
I guess just label this as a feature then?

Nebyab:
i mean sure alright

ghost:
good shit tbh. Also maybe the damage dealt should be changed so that it's not static in case the ass-damage is ever changed

Nokington:
:+1:

Issue #2003, opened 2016-03-13
Disposals mapping problems


Spacedong:
https://vgy.me/iucWsY.png

the shutter button and mass driver button cannot be reached because they are covered by the window

the interior door leading to the cargo chute is always open and its sprite can not be clicked on

Issue #2009, opened 2016-03-16
SecTech in Armory has bad placement


Spacedong:
https://vgy.me/TG6Jc2.png

I know you can kick the rack to get to it but its just bad placement all around.

trifornt-tinfornt:
Kilm

Pyko1:
I kinda miss my sec before it was pissed all over.

Spacedong:
kilm

carlarc:
you can still get the items from it, dunno why it's "bad placement". you can click on it, you can vend items, you can pick them up. the rack doesn't do shit

dragonrobot:
It looks like ass

Pyko1:
fixed

PR #2021, opened 2016-03-18
Mapfix

Pyko1:
Fixed disposals buttons and moved things around the armory.
Added 2 barriers.
capture

Fixes: https://github.com/HippieStationCode/HippieStation13/issues/2003 https://github.com/HippieStationCode/HippieStation13/issues/2009


Nebyab:
ok

carlarc:
Much easier to get to the loyalty implants, A little harder for the lawyer to just run in and tase the warden. Though I'd like for the empty spot above to have a rack or a table or something.

Pyko1:
I actually used to have rack there, but removed it since it made the corner table look shit.
A table could do, but getting over one or dismantling it can be a real pain in the ass with airflow, hence I decided to leave an empty spot if anyone wants to break in from space. Half tempted to move the table for same reason.

carlarc:
thats kind of the reason for me to want it tho

Nokington:
:+1:

Issue #2022, opened 2016-03-18
Hacking Book Leads to 404 Error


Zadeon:
http://i.imgur.com/jvETOOu.png?1

What is the page it's suppose to link to?

Spacedong:
that_fukken_gator.png

PR #2023, opened 2016-03-18
Guidebook Fix?

AliiTea:
This might fix guidebooks being broken..

Fixes: https://github.com/HippieStationCode/HippieStation13/issues/2022
_maybe_

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -59,10 +59,10 @@
 
 	var/server
 	var/banappeals
-	var/wikiurl = "http://www.tgstation13.org/wiki" // Default wiki link.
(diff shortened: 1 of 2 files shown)


Carbonhell:
No. Those values get changed to the config values,those one you edited are the default values so they don't affect that bug, to fix the bg just go check where hippie-station.com is used and change it with pur new host hippiestation.com

TDcoolguy300:
As carbon said, these are primarily default values and are overwritten once the configurations have loaded

Carbonhell:
Just close your PR and make a new one when you fix your code

AliiTea:
Aye aye

PR #2026, opened 2016-03-18
makes the proxycheck return value a num

Carbonhell:
I'm an idiot,this should FINALLY fix every single issue about proxycheck


Pyko1:
code looks fine and apparently a bug fix, merging

PR #2032, opened 2016-03-21
Shadowling Proposed nerfs

Kokojos:
Here's what this PR does to shadowlings at the current time.

Most skill cooldown increased.
Glare is now reduced by 3/10th of it's time by having eyewear.
Shadow Walk is now much, much faster, but leaves a trail of shadows behind it.
Veil does not affect glowshrooms anymore.
Blindness smoke stuns less
Sonic Screech confusion time reduced by 2/10th.
Icy veins now does not affect people with suits, but deals a little more damage overall.
Shadowlings can't use any abilities while Shadow Walking.(Bugfix)
Thrall lesser glare now has a range of 2.

All of those nerfs are up for debate execpt the icy veins ones and using skill while another dimension.

 I do not claim to have ultimate knowledge, but I do know, as a veteran player, that I never lost as a Shadowling. I feel those nerfs are not extreme, just enough to make the Sling lose their ''edge'' that they have with their incredible variety. Being able to kill a Sling 1v1 is still impossible (unless the Sling's an idiot) but this gives more chances to people working as a Team.

shadownerf1

### DIFF
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -1,6 +1,9 @@
 #define EMPOWERED_THRALL_LIMIT 5
 
 /obj/effect/proc_holder/spell/proc/shadowling_check(var/mob/living/carbon/human/H)
+	if(H.incorporeal_move == 2)
+		usr << "<span class='warning'>You cannot do this while in another dimension!</span>"
+		return
 	if(!H || !istype(H)) return
 	if(H.dna.species.id == "shadowling" && is_shadow(H)) return 1
 	if(H.dna.species.id == "l_shadowling" && is_thrall(H)) return 1
@@ -14,7 +17,7 @@
 	name = "Glare"
 	desc = "Stuns and mutes a target for a decent duration."
 	panel = "Shadowling Abilities"
-	charge_max = 300
+	charge_max = 350
 	human_req = 1
 	clothes_req = 0
 	action_icon_state = "glare"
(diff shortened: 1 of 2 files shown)


trifornt-tinfornt:
Icy Veins is overpowered but making it so that anyone wearing a suit (do you just mean hardsuits, biosuits, or just anything in the exosuit slot? i believe it's the very latter) is an extremely dumb nerf and isn't going to fix anything besides making Head of Security powergamers and security in general way stronger against shadowlings - especially thanks to their glasses.

Otherwise, yeah, I agree with most of these nerfs. Shadowling should not be as powerful as it is and I always wondered why wearing glasses didn't reduce stuns from glare.

Kokojos:
Only space-worhty suits.

trifornt-tinfornt:
Okay, that's great. Push it.

Actually on second thought why does Shadow Walk leave a trail? That seems kinda against the idea of the semi-stealthy nature of Shadowlings. One guy shadow walking down the hall is gonna get them immediately called out.

code/game/gamemodes/shadowling/shadowling_abilities.dm
```diff
@@ -136,11 +140,12 @@
 	desc = "Instantly freezes the blood of nearby people, stunning them and causing burn damage."
 	panel = "Shadowling Abilities"
 	range = 5
-	charge_max = 250
+	charge_max = 350
 	human_req = 1
 	clothes_req = 0
 	action_icon_state = "icy_veins"
 	sound = 'sound/effects/ghost2.ogg'
+	var/suits = list(/obj/item/clothing/suit/space) //Apparently dosen't work if not in a list?
```
TDcoolguy300, line 148: You could create a temp var in the proc itself rather than defining it up here

WhiteFox08:
Shadowling is by itself already pretty difficult to play, i mean sure veterans will be very overpowered as shadowlings. However, this would make a already difficult antag even more difficult for those who are not as experienced. Personally I wouldn't recommend this PR.

code/game/gamemodes/shadowling/shadowling_abilities.dm
```diff
@@ -148,20 +153,21 @@
...
 					continue
-			M << "<span class='userdanger'>A wave of shockingly cold air engulfs you!</span>"
-			M.Stun(2)
-			M.apply_damage(10, BURN)
-			if(M.bodytemperature)
-				M.bodytemperature -= 200 //Extreme amount of initial cold
-			if(M.reagents)
-				M.reagents.add_reagent("frostoil", 15) //Half of a cryosting
+			if(istype(M.wear_suit))
+				if(M.wear_suit.type in suits)
```
Carbonhell, line 164: Instead of this snowflakey check with that suits list,keep in mind that clothing has a min_cold_protection_temperature var, which spacesuits have it set to min_cold_protection_temperature = SPACE_SUIT_MIN_TEMP_PROTECT . Check for this value since it handles cold protection

Nokington:
Not sure about the shadowwalk thing

Kokojos:
Aye, I think Il remove the shadowalk thing. I went a bit too far in my toughts.

Kokojos:
Bareboning this in another PR.

PR #2034, opened 2016-03-21
Increases the Autodrobe Refiller size.

TheOneAndOnlyCreeperJoe:
Autodrobe contains 81 standard items, refiller was outdated and as such the Jester Hat, Corgi Suit and Carp Suit would be physically impossible to acquire without admin fuckery (since you can only fit 3 refillers into one autodrobe.)

Increased the amount by 2 (previously 25 now 27), allowing those 3 items to be picked.


Nokington:
:+1:

Carbonhell:
Code's fine and it's a bugfix,merging

PR #2041, opened 2016-03-22
fixed misspellings with moth wings

### LINKED ISSUE #2038: Mothmen wingstyle "Lovers" are broken.
Hi there seems like the Lovers wing style is broken, in the code the name of the wingstyle in the dmi is called lovers while the code itself seems to point to icon_state = Lovers as seen below.

`icon_state = "Lovers"`

the icon_state is written with a capital L 

simply changing it to 
`icon_state = "lovers"`

would fix this.

Update the Redish wingstyle is broken aswell the problem here is because the file name for a part of the wing is called `moth_m_wing_radish_2` while it should be `moth_m_wing_redish_2` this is a mistake on my part.

Spacedong:
fixes #2038


Carbonhell:
Code ok,merging

TDcoolguy300:
I was exhausted when i did that, RIP

PR #2046, opened 2016-03-23
Anti-stun update quickfix

Kierany9:
Quick fix to a couple of things I managed to mess up with my last PR that nobody caught. Adds a 5-cycle(10 in the case of Ephedrine) minimum for the stuns of these chemicals to kick in, as adding a very low dose to a syringe gun would provide an instant 5-second stun as well as heavy stamina damage. Also changed the stamina damage a bit for balancing purposes. Crank now does 20 Stamina damage and Meth does 35.


Carbonhell:
code's fine and it's a bugfix,merging

Issue #1879, opened 2016-02-15
Engieborgs can only pick up red wire.


Spacedong:
Trying to pick up any other color wire with your wire stack tool does nothing. It can only be refilled with red wire. What the fuck?

McBawbaggings:
Can confirm. Red coils worked and the blue one didn't.

JustinBNewton:
Same with any other color of wire. Not a bug.

Spacedong:
what? what are you talking about, this is a bug, you can pick up other wire colors as human, but not a borg.

JustinBNewton:
Two ways to fix this if you really wanted to, 
1. Add every color of wire to the borg.
2. Make a wire that is able to switch colors.

Spacedong:
the borg's wire CAN switch color you dumbo

Spacedong:
STILL doesn't work

JustinBNewton:
Then switching wire color as a borg doesn't work?

JustinBNewton:
If you click any other color of wire with a red wire, it just makes it all become red. Thats a feature.

Spacedong:
I think you're missing the point here, you can't pick up any other wire color on the ground as a borg except red.

JustinBNewton:
Ah gotcha, thanks for clarifying.

Grufus2:
One solution I could see is to add something like this:

```
cable.dm:line 493

/obj/item/stack/cable_coil/attackby(obj/item/W, mob/user, params)
	..()
	if(istype(W,/obj/item/stack/cable_coil/cyborg))
		var/obj/item/stack/S = W
		merge(S)
		user << "<span class='notice'>Your [S.name] stack now contains [S.get_amount()] [S.singular_name]\s.</span>"
```

Pyko1:
might actually look in to this and slap it on my noose & gravity PR since i'm working on the file anyways,

Pyko1:
It seems engi borgs dont exactly use amount on coils, instead they have their own var for cable synthetiser, which seems to be only able to convert the red cable that is parent of all coils (colored and random). Also while snooping around I noticed it had recharge value but it does not seem to recharge at all on its own. Was not actually able to find the code that takes care of converting cable amount in to cable synth charge.

McBawbaggings:
Engiborg's cable synth is only recharged in cyborg docks
I think it's this proc that handles the recharging of modules like the cable coil:
[...]

PR #2047, opened 2016-03-23
Fixed stacking cables of different types

JohnGinnane:
After a little testing it seems neither humans nor cyborgs could stack a
"random" cable (/obj/item/stack/cable_coil/random/) onto a normal type
(/obj/item/stack/cable_coil) or cyborg table
(/obj/item/stack/cable_coil/cyborg). I've made it so it checks if both
the target and the weapon are the same "type" in another method.

I've also added a little message to let the non-robot know that the
stack is already at maximum capacity when they try to add something to
the stack. (e.g. 30 units for cable coils)


JohnGinnane:
This is a fix for https://github.com/HippieStationCode/HippieStation13/issues/1879

code/game/objects/items/stacks/stack.dm
```diff
@@ -236,8 +236,13 @@
 	return
 
 /obj/item/stack/attackby(obj/item/W, mob/user, params)
-	if(istype(W, src.type))
+	if(istype(W, src.type) || (istype(W, /obj/item/stack/cable_coil) && istype(src, /obj/item/stack/cable_coil)))
```
JohnGinnane: It seems as though if W type was `/obj/item/stack/cable_coil` or `/obj/item/stack/cable_coil/cyborg` and src type was `/obj/item/stack/cable_coil/random` it wouldn't match

code/game/objects/items/stacks/stack.dm
```diff
@@ -236,8 +236,13 @@
...
 		var/obj/item/stack/S = W
+
+		if(S.amount >= S.max_amount && !isrobot(user))
```
JohnGinnane, line 242: Added a nice little message to let the non-robot user know if their stack was already full

Carbonhell:
I don't really like that snowflake check just for cables. There's a variable called parent type which is the path of the parent,for example parent type of /obj/item/stack/cable_coil/random/ would be /obj/item/stack/cable_coil . maybe you could use this ?Or perhaps just scratch the random type of cables and make it be inside the normal cable_coil new proc?

JohnGinnane:
Sounds good, I'll look into using the parent type instead.

As for the random type of cables, how would you go about implementing it into a new procedure?

myontasm:
Fix the snowflake code and I'll give it a +1

JohnGinnane:
Okay I'll take care of that

Carbonhell:
Actually using parent type might fuck up with all other stack types. The best way would be removing the cable coil random types

JohnGinnane:
Can you tell me where the maintenance loot is generated and I can take care of it there?

JohnGinnane:
Okay I'll look at this tonight

code/modules/power/cable.dm
```diff
@@ -484,8 +484,13 @@ var/global/list/datum/stack_recipe/cable_coil_recipes = list ( \
 /obj/item/stack/cable_coil/New(loc, amount = MAXCOIL, var/param_color = null)
 	..()
 	src.amount = amount
+
 	if(param_color)
 		item_color = param_color
+	else
+		item_color = pick("red","yellow","green","blue","pink")
```
JohnGinnane, line 491: If no colour is specified on cable_coil creation it will pick one at random

JohnGinnane:
Okay dokey, that should be that done. There's no longer  "random" type cable coil. Instead it's random if you create it without specifying the colour.

myontasm:
Good stuff :+1:

Carbonhell:
code's ok,merging

PR #2054, opened 2016-03-25
Allows self-targeting with staves.

TheOneAndOnlyCreeperJoe:
Wizards can now shoot themselves with most staves (besides animation and door) to apply the effects to themselves. Most notable use is using the Staff of Healing on yourself without having to stick it in your mouth first.

Chaos staff is probably incredibly shoddy (but at least doesn't cause any errors/warnings)

Uses recycled wand code.

### DIFF
--- a/code/modules/projectiles/guns/magic/staff.dm
+++ b/code/modules/projectiles/guns/magic/staff.dm
@@ -1,5 +1,32 @@
 /obj/item/weapon/gun/magic/staff/
 	slot_flags = SLOT_BACK
+	
+/obj/item/weapon/gun/magic/staff/attack(atom/target, mob/living/user)
+	if(target == user)
+		return
+	..()
+
+/obj/item/weapon/gun/magic/staff/afterattack(atom/target, mob/living/user)
+	if(!charges)
+		shoot_with_empty_chamber(user)
+		return
+	if(target == user)
+		if(no_den_usage)
+			var/area/A = get_area(user)
+			if(istype(A, /area/wizard_station))
+				user << "<span class='warning'>You know better than to violate the security of The Den, best wait until you leave to use [src].<span>"
+				return
+			else
+				no_den_usage = 0
+		zap_self(user)
+	else
+		..()
+	update_icon()
+	
+/obj/item/weapon/gun/magic/staff/proc/zap_self(mob/living/user)
+	user.visible_message("<span class='danger'>[user] zaps \himself with [src].</span>")
+	playsound(user, fire_sound, 50, 1)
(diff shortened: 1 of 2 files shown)


Carbonhell:
Iunno, in my opinion shooting yourself with a staff should require a delay to avoid selfhealing yourself. The huge delay shooting in your own mouth has fits perfectly and i don't think this PR is really needed, since the only staff a wizard will want to shoot himself with is the healing one, and in very rare circumstances the staff of change.

myontasm:
I agree with Carbon on the delay thing.

TheOneAndOnlyCreeperJoe:
@carbonhell the problem is is that nobody picks the staff of healing because the wand from the wand assortment do allow you to instantly heal yourself. Combined with charge and you have ~35 charges at maximum on your wand of healing (more than you'll ever need) that can instantly heal yourself as well as 4 other wands that you can use.

Carbonhell:
Personally i prefer the staff of healing,because it has no limit, and buying wands + charge equals to 3 o 4 spell points,whereas only the staff of healing costs 1(or 2?unsure) points. I always buy only the staff of healing for this reason, and you can use it to heal other people without worrying about losing a charge

trifornt-tinfornt:
I think there should be a brief delay on the staff of healing shooting yourself but it shouldn't be as long as sticking it in your mouth. The only reason most people legitimately buy the Wand Assortment is for the wand of healing.

Carbonhell:
um what no that shit gives the wand of death which is an one hit kill too, all the wands are pretty useful to be fair, and i'm fine with the delay it currently has,it's my opinion through

ghost:
sticking it in your mouth was perfect balance
if interrupted you fuck up, and it takes time
no need to buff it

Carbonhell:
some people just like to stick it in their mouth, we can't help it

ghost:
If this does get merged, you're gonna need a changelog

ghost:
Out of curiosity, can someone link me the code that handles putting staves in your mouth? I can't find anything to suggest the mechanic in the source code.

Nokington:
I dunno how I feel about this. I don't really think its a good idea.

Nebyab:
Ehh. The balance of staffs and staves is that they're difficult to use on yourself. I think the current implementation is utterly fine

blobbernaut:
not necessary or a smart balance choice

Crystalwarrior:
I don't think this is that good of an idea honestly

Carbonhell:
All headmins vouched in and noone liked this. Closing it.

PR #2055, opened 2016-03-25
Fixes the Bottle of Blood

### LINKED ISSUE #2039: Wizard Bottle of Blood is broken.
What it says on the tin. It has no sprite, it has no name and activating it does nothing. Spawned it separately from both the admin panel and from the wizard spell book.

Bottle of Blood is the Slaughter Demon spawner by the way.

Also, refunding the Bottle of Blood yields only 1 point in refunds, but costs 2 points to buy. (This affects   the Scroll of Apprenticeship as well)

TheOneAndOnlyCreeperJoe:
Fixes #2039

Better worded as added as it didn't exist in the code before.


code/game/gamemodes/antag_spawner.dm
```diff
@@ -205,7 +205,36 @@
...
+	if(used)
+		user << "<span class='warning'>Somehow the bottle has already been emptied!</span>" // shouldn't happen as it destroys itself upon succesful summoning
+		return 0
+	if(user.z == ZLEVEL_CENTCOM)
+		user << "<span class='warning'>There are no demons within this realm. Try using the [src] onboard the station.<span>"
+		return
+	return 1
 
+/obj/item/weapon/antag_spawner/slaughter_demon/attack_self(mob/user)
+	if(!(check_usability(user)))
```
Carbonhell: You should not use ! here, otherwise it'll do the exact opposite aka fail when you can use it, and succeed when you can't.

code/game/gamemodes/antag_spawner.dm
```diff
@@ -205,7 +205,36 @@
...
+	icon = 'icons/obj/wizard.dmi'
+	icon_state = "vial"
+	
+/obj/item/weapon/antag_spawner/slaughter_demon/proc/check_usability(mob/user)
+	if(used)
+		user << "<span class='warning'>Somehow the bottle has already been emptied!</span>" // shouldn't happen as it destroys itself upon succesful summoning
+		return 0
+	if(user.z == ZLEVEL_CENTCOM)
+		user << "<span class='warning'>There are no demons within this realm. Try using the [src] onboard the station.<span>"
+		return
```
Carbonhell: return 0 rather

code/game/gamemodes/antag_spawner.dm
```diff
@@ -205,7 +205,36 @@
...
+
+	var/list/wizard_candidates = get_candidates(ROLE_WIZARD, 3000, "slaughter demon") // best pref for slaughter demon, as it isn't a pref on it's own.
+	if(wizard_candidates.len > 0)
+		used = 1
+		var/client/C = pick(wizard_candidates)
+		spawn_antag(C, get_turf(loc), "Slaughter Demon")
+		new /obj/effect/gibspawner/human(loc)
+		playsound(loc, 'sound/effects/Glassbr2.ogg', 25, 1)
+		user << "<font size=3><span class='danger'><b>\The bottle sizzles and the bottle shatters as the contents splatter over the floor, only to burn away moments later. You feel a sense of dread wash over you as the contents dissapear without trace.\</font></span>"
```
Carbonhell: the bottle sizzles and shatters, rather

code/game/gamemodes/antag_spawner.dm
```diff
@@ -205,7 +205,36 @@
...
+		qdel(src)
+	else
+		user << "<span class='warning'>No demons were attracted to the blood. You can either wait, or refund the bottle by using it on your book</span>"
```
Carbonhell: missing a final dot

code/game/gamemodes/antag_spawner.dm
Carbonhell: also there's a \ at the end which isn't needed apparently. and maybe you'd like this message to be displayed to everyone on sight so visible_message() instead of user <<

code/game/gamemodes/antag_spawner.dm
```diff
@@ -205,7 +205,36 @@
...
+/obj/item/weapon/antag_spawner/slaughter_demon/attack_self(mob/user)
+	if(!(check_usability(user)))
+		return
+
+	var/list/wizard_candidates = get_candidates(ROLE_WIZARD, 3000, "slaughter demon") // best pref for slaughter demon, as it isn't a pref on it's own.
+	if(wizard_candidates.len > 0)
+		used = 1
+		var/client/C = pick(wizard_candidates)
+		spawn_antag(C, get_turf(loc), "Slaughter Demon")
+		new /obj/effect/gibspawner/human(loc)
```
Carbonhell: loc will probably make the gibs spawn in the human. instead of loc, use get_turf(src)

code/game/gamemodes/antag_spawner.dm
```diff
@@ -205,7 +205,36 @@
 
 
 ///////////SLAUGHTER DEMON
+/obj/item/weapon/antag_spawner/slaughter_demon
+	name = "Bottle of Blood"
```
Carbonhell: names should not be capsed,because when examined it'll say "This is the Bottle of Blood."

code/game/gamemodes/antag_spawner.dm
```diff
@@ -205,7 +205,36 @@
...
+	if(!(check_usability(user)))
+		return
+
+	var/list/wizard_candidates = get_candidates(ROLE_WIZARD, 3000, "slaughter demon") // best pref for slaughter demon, as it isn't a pref on it's own.
+	if(wizard_candidates.len > 0)
+		used = 1
+		var/client/C = pick(wizard_candidates)
+		spawn_antag(C, get_turf(loc), "Slaughter Demon")
+		new /obj/effect/gibspawner/human(loc)
+		playsound(loc, 'sound/effects/Glassbr2.ogg', 25, 1)
```
Carbonhell, line 232: not sure if loc is what you want here

Carbonhell:
good to go,it's a bugfix

PR #2062, opened 2016-03-26
Fixed Wheelchair

### LINKED ISSUE #2060: Wheelchairs
Wheelschairs are moving backwards now.
Also they seem to be slow as fuck.

JohnGinnane:
I've fixed the wheel chair facing the wrong direction using what I think might be snowflake code. However considering that's the only wheeled object (that I know of) that faces the correct direction when you try to move on it I think it's okay

I've also improved the speed as suggested

Fixes #2060


JohnGinnane:
As reported: https://github.com/HippieStationCode/HippieStation13/issues/2060

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -147,27 +147,27 @@
...
-		switch(buckled_mob.dir)
-			if(NORTH)
-				buckled_mob.dir = SOUTH
-			if(WEST)
-				buckled_mob.dir = EAST
-			if(SOUTH)
-				buckled_mob.dir = NORTH
-			if(EAST)
-				buckled_mob.dir = WEST
+		if(!istype(src, /obj/structure/bed/chair/withwheels/wheelchair))
```
JohnGinnane: @Carbonhell you might let me know if this is too "snowflake". I'm not sure if there's any other objects like the wheelchair.

I was just thinking we could have a flag/variable on the object saying whether if should face forward/backward when moving, which would allow new objects to use this feature without having to change this line specifically. What are your thoughts?

Carbonhell:
Yeah no. This is way too snowflakey and will only lead to issue. Rather, you could make two new arguments to the proc, delay and direction, and handle it like that so each object can set them separately easily.

JohnGinnane:
That's what I was thinking, thanks for the guidance

Carbonhell:
For future reference, in your PRs in the description write "Fixes #NUMBER" to let the issue automatically close when your PR gets merged.

JohnGinnane:
Ah, cheers!

JohnGinnane:
How's this @Carbonhell

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -139,40 +139,44 @@
...
+				if(SOUTH)
+					buckled_mob.dir = NORTH
+				if(EAST)
+					buckled_mob.dir = WEST
 		dir = buckled_mob.dir
 	handle_rotation()
 	handle_layer()
 	cooldown = 1
-	spawn(10)
+	spawn(src.delay ? src.delay : delay)
```
Carbonhell: you've set a default value to all "withwheels" chairs, var/delay = 10. If your child won't have a different value,it'll inherit the parent's one,therefore you can just do spawn(delay)

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -139,40 +139,44 @@
 	color = rgb(255,251,0)
 
 /obj/structure/bed/chair/withwheels
+	var/delay = 10
 	var/cooldown = 0
+	var/reverse_direction = 1
 
-/obj/structure/bed/chair/withwheels/relaymove(mob/user, direction)
+/obj/structure/bed/chair/withwheels/relaymove(mob/user, direction, delay=10)
```
Carbonhell: This doesn't need a delay argument, you can just use the var/delay = 10 var

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -139,40 +139,44 @@
...
-			if(NORTH)
-				buckled_mob.dir = SOUTH
-			if(WEST)
-				buckled_mob.dir = EAST
-			if(SOUTH)
-				buckled_mob.dir = NORTH
-			if(EAST)
-				buckled_mob.dir = WEST
+		buckled_mob.dir = direction
+		if(src.reverse_direction)
```
Carbonhell: using src. is most of the time useless, unless you have a case of an argument having the same name of a variable(like your current delay var/arg), just saying. It'd be better to remove that src.

Carbonhell:
Good job,just fix those small issues and it'll be good to go. Also, you've been coding a lot and you are pretty competent aswell, you may want to join our slack channel(um..is this a deja vu or did i already say this?i'm becoming retarded)

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
JohnGinnane: Is "src" is the equivalent of "this" in things like C#? If so that makes sense

JohnGinnane:
Done, and yes I've pestered you a bit on getting invited to Slack!

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -139,40 +139,44 @@
...
+				if(SOUTH)
+					buckled_mob.dir = NORTH
+				if(EAST)
+					buckled_mob.dir = WEST
 		dir = buckled_mob.dir
 	handle_rotation()
 	handle_layer()
 	cooldown = 1
-	spawn(10)
+	spawn(delay ? delay : 10)
```
Carbonhell: why did you leave "delay ? delay : 10"? The default value is already 10 mate

JohnGinnane: Whoops, sorry, should be fixed now

Carbonhell:
Perfect!Thank you,if limey knows how to invite people on slack you should be invited in the next minutes

JohnGinnane:
Thanks!

PR #2065, opened 2016-03-27
Fixed washing machine not updating clothes' names if the colour was changed

### LINKED ISSUE #1279: coloring gloves problem
tryed to paint my yellow gloves with mimes grayon, but it didn't work

JohnGinnane:
Most colours will be updated to match the colour of the crayon/stamp
placed in the washing machine. However the descriptions remain unchanged.

Fixes #1279

### DIFF
--- a/code/game/machinery/washing_machine.dm
+++ b/code/game/machinery/washing_machine.dm
@@ -65,13 +65,12 @@
 			for(var/obj/item/clothing/I in contents)
 				I.color = wash_color //Simply recolor the items.
 
-				var/colors = list("light brown", "brown", "cyan", "fingerless", "combat", "tactical", "yellowgreen", "darkred", "lightred", "maroon", "red", "orange", "rainbow", "lightgreen", "green", "lightpurple", "purple", "gold", "darkblue", "lightblue", "aqua", "blue", "yellow", "black", "grey", "gray", "white", "latex", "nitrile", "budget insulated", "insulated", "captain's")
+				var/list/colors = list("light brown", "brown", "cyan", "fingerless", "combat", "tactical", "yellowgreen", "darkred", "lightred", "maroon", "red", "orange", "rainbow", "lightgreen", "green", "lightpurple", "purple", "gold", "darkblue", "lightblue", "aqua", "blue", "yellow", "black", "grey", "gray", "white", "latex", "nitrile", "budget insulated", "insulated", "captain's")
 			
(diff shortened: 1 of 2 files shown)


Nokington:
k

code/game/machinery/washing_machine.dm
```diff
@@ -65,6 +65,13 @@
 			for(var/obj/item/clothing/I in contents)
 				I.color = wash_color //Simply recolor the items.
 
+				var/colors = list("light brown", "brown", "yellowgreen", "darkred", "lightred", "maroon", "red", "orange", "rainbow", "lightgreen", "green", "lightpurple", "purple", "gold", "darkblue", "lightblue", "aqua", "blue", "yellow", "black", "grey", "gray", "white", "latex", "nitrile", "budget insulated", "insulated", "captain's")
+				
+				for(var/old_color in colors)
+					if(old_color != wash_color)
```
Carbonhell: I don't get this,i doubt it can work anyhow

JohnGinnane: I'm looping through a list of common colours used for various clothes like jumpsuits, hats, gloves, shoes, bandanas, etc.

Then I just replace that word with the wash color. On the one hand it an item's name contains more than one colour in it's name they'll both get changed to wash_color, but on the other hand I don't see very many of those items.

I guess the if statement is redundant, but it's basically excluding the potential one case where the old_color matches the wash_color

JohnGinnane:
Here's an example screenshot:

image

doobiedoo23:
So if I get tactical gloves or something and throw them in a washing machine with a crayon, would the name change to '(color of crayon) gloves'?

JohnGinnane:
If the name is "tactical" not yet, but I can add that to the list! Currently discussing other ways to change the name

JohnGinnane:
I've added "combat", and "tactical" to the list of words to update. I've also removed the icon_state being updated because you could end up with results like this (this is actually a bandana):

image

Carbonhell:
Due to it changing capn gloves, tactical/fingerless/etc stuff,it's a feature now so i need a second head ok before merging
code's ok tho

doobiedoo23:
Opens up a few more opportunities to traitors, which is nice. +1

Nebyab:
sure

Kokojos:
two heads :+1: , merging

PR #2073, opened 2016-03-28
Blacklists (most) exotic species from the Magic Mirror

TheOneAndOnlyCreeperJoe:
Shadowlings, Thralls, Shadowpeople, Podpeople, Plantpeople, Adamantite Golems, Slimes, Skeletons (round start and exotic variant), Golems, Meeseeks, Zombies, Abductors and Plasamen will no longer be available trough the Magic Mirror, as they caused certain unintentional gameplay effects (broken OP armor to the point of immortality, Abductor telepathic link with apprentices, health regen with podpeople) that would normally be an intentional weakness for the wizard.

This does not affect the badmin version of the mirror.

**Shadowlings, Thralls and Shadowpeople** were blacklisted because they only possess the darkness mechanic, their other abilities are unavailable and you will basically burn up before you can reach a maintenance shaft. In addition, they do not need to breathe, are immune to radiation and cannot bleed.
**Pod/Plantpeople** have health regen in light (1 for each damage type per second) with the only downside being that you get more 'hungry' in the dark.
**(Adamantite) Golems** have the armor exploit (not going to disclose details or the next few wizard rounds will be hell), as well as a natural hardsuit that normally cannot be removed, immunity to blood loss and are immune to to radiation. In addition, their natural hardsuit has 60% protection from melee, lasers, stamina, thrown items and bullets (although they are slowed down by it) as well as blocking syringes and stops items from embedding.
**Abductors** allowed for telepathy between apprentices with virtually no drawback (talking is almost irrelevant as a Wizard anyway if you are just out for murder).
**Zombies and Slimepeople** gave immunity to bleeding, as well as Zombie having temperature resistance, no need to breathe and immunity to radiation.
**Meeseeks and Plasmamen** are just stipid. Plasmamen instantly combust the moment you swap with them and permanently leaves an error on the screen if you change races. Meeseeks just instantly turns you back to human anyway.
**Skeletons** were blacklisted because the exotic one is broken OP and only available to lichdom wizards, normal skeletons are a child and therefore must be banned alongside it.

With the current state of the mirror, **there is no reason to play anything other than an exotic species** (unless you are going mutate since it's human exclusive) since they are almost guaranteed to be better.


myontasm:
That's a lot and I'm not sure I like this.

doobiedoo23:
Just removing skeletons and golems would be fine.

Crystalwarrior:
@doobiedoo23 I think "broken" races should go too, like Meseeks and Plasmamen. At the very least make Plasmamen wizards have their own special equipment as soon as you change into them.

Kokojos:
It's extremely easy to set species on the side if we want admin fuckery, and this prevent a lot of Powergamer opportunities and broken mechanics. :+1:

Nokington:
Yes

Nebyab:
This is something i've complained about for a while, and only gotten worse as of late

Carbonhell:
um nebyab so do you fully agree with it,and what about limey disagreeing

TheOneAndOnlyCreeperJoe:
@Nebyab @Limed00d

myontasm:
I've said I disagree with it mostly because of choosing to remove rather than to fix most exploits.

blobbernaut:
i absolutely disagree with this change, pod people, abductors, slimes, shadow people, are all not OP enough to remove from the magic mirror. They're fucking WIZARDS. Oh no, so they heal 1 health per tick while in the light. This should be an option available to them.

Remove only the ones that actually are broken, codewise, and the ones that give you space resistance (zombie, skeleton, golem). To the other removals, no

blobbernaut:
I don't think you realize that being able to gain these special powers was the whole point of the magic mirror

Nebyab:
I agree with it. The point of the magic mirror is to customize who you are, not to give yourself whatever species you want, most of which are tethered to direct buffs which the wizman shouldn't have.

Pyko1:
Seems like we have 2 for and 1 against then, so approved I guess?

Nebyab:
I don't believe this has been codechecked, so its not merge ready.

PR #2078, opened 2016-03-28
Added speed boost upgrade for Beepsky and medibots

JohnGinnane:
This boost only lasts 5 seconds, but will make him move at 250% of the original speed. This boost has a cooldown of 30 seconds.

This makes straight dashes away from a Beepsky/Pingsky more difficult


Kokojos:
There's nothing but a changelog?

JohnGinnane:
Whoops

Spacedong:
yes we need this to catch methheads

JohnGinnane:
It's not that fast I don't think, but you'll have a very hard time outrunning it in a straight hallway

carlarc:
**noooooo**

Kierany9:
What the fuck, I proposed this as an april fools update, not a real thing.

JohnGinnane:
So you don't like it?

Kierany9:
Nah, it's alright. No idea how balanced it would be but it seems ok.

Crystalwarrior:
beepskies are fun police 99% of the time
I remember when I got beepski'd for no reason other than having a gas mask on (unknown identity bullshit), HoS comes, fullstrips me, finds traitor gear and executes me on the spot.

code/game/machinery/bots/secbot.dm
```diff
@@ -242,7 +246,7 @@ Auto Patrol: []"},
 
 				else								// not next to perp
 					var/turf/olddist = get_dist(src, target)
-					walk_to(src, target,1,4)
+					walk_to(src, target, 1, 4 * (fast == 1 ? fast_delaymod : 1))
```
Carbonhell: shouldn't it be (fast == 1 ? fast_delaymod : 4) since the previous value was 4?

code/game/machinery/bots/secbot.dm
```diff
@@ -349,6 +353,13 @@ Auto Patrol: []"},
 			mode = BOT_HUNT
 			spawn(0)
 				bot_process()	// ensure bot quickly responds to a perp
+			if(fast == 0)
```
Carbonhell: if(!fast)

code/game/machinery/bots/secbot.dm
```diff
@@ -349,6 +353,13 @@ Auto Patrol: []"},
...
+				fast = 1
+				spawn(fast_length * 10)
```
Carbonhell: have you tested it?it doesn't stop all the other stuff right?

html/changelogs/JohnGinnane - buffBeepsky.yml
```diff
@@ -0,0 +1,6 @@
+author: John Ginnae
```
Carbonhell: You spelt your name wrong ayy :^)

Carbonhell:
iunno about the code,i'll check it again when john is available for questions

code/game/machinery/bots/secbot.dm
JohnGinnane: The delaymod variable is a modifer, so 0.4 would mean the delay is only 40% usual. I can replace it with a "boostdelay" and give it a value of 1.6 which is the same result

code/game/machinery/bots/secbot.dm
JohnGinnane: Noted

code/game/machinery/bots/secbot.dm
JohnGinnane: I have tested this on a local copy, and it works correctly. What's more is that if you attack Beepsky he will chase you as usual but not use this ability, it's only when he finds a player whose set to arrest.

Also from what I've read about "spawn" it basically places the scope in some sort of location that runs after the amount of time specified right? So the rest of the code can execute and then after the specified time the other code will run. Not sure exactly how this works but the result is what I expect

html/changelogs/JohnGinnane - buffBeepsky.yml
JohnGinnane: Cheers

JohnGinnane:
I'll do some revisions now, but just a quick thing, should I not move the entire thing into it's own proc such as "activateBoost" and "reset" etc.? It seems like a more OOP friendly approach. What do you think?

Nebyab:
I love beepsky to death.

I mean that unironically.

This is a terrible idea. Beepsky has two disadvantages: He's slow and he can't put people in the brig. Giving him superspeed seems bad, personally, since now Beepsky is better than the average sec officer in virtually every way, with no reason for it.

JohnGinnane:
Can anything be done to balance or are you thinking scrap the whole thing?

Pyko1:
Maybe we could have this as research item, Secbot rocketbooster or something.

JohnGinnane:
I don't think anybody would invest time in researching it, would hacking it using a multitool be easier/more preferable?

Pyko1:
I think having it behind mere hacking would not really limit it that much.

If it was in science it gets researched in its own if anyone does research, then its just matter of getting the mats and building the upgrade.

Nebyab:
As a default beepsky buff, I see no way for this to work.

As an upgrade, research project, or something like that it would be alright.

JohnGinnane:
So would the upgrade be a whole new beepsky or a component in the materials for beepsky so new units are made with the ability?

JohnGinnane:
@Nebyab @Pyko1 what are thoughts and making this something medibots can use too? Is that too far? As a generic "speed boost" circuit you could install

Pyko1:
I see no harm in having it even on cleanbots.

JohnGinnane:
Oh, it's a temporary boost, only 5 seconds (right now). After that it's back to normal speed for 30 seconds

JohnGinnane:
What if we make it a constant increase for floorbots and medibots and then a much higher but temporary one for securitrons? (5 second length, 30 second cooldown)

JohnGinnane:
Hi all, I've completely revamped the code so it's much more friendly to make changes to and allows us to potentially add other upgrades for other bots (I have some ideas in mind!)

Pyko and Neb, at the moment the way it works is when securitron first comes across a criminal it will use it's boost ability to quickly dash towards them. This lasts 5 seconds and once used will then have a 30 second cool down before being able to use it again. 

If we were to add this upgrade card to other bots (and I would love to do that) then in other bots it would mean the upgrade would result in a constant speed bonus, maybe 150% of default, instead of 250%. On the other hand we could just make securitron faster in general but the sudden dash is intimidating and more creative than just a straight upgrade in speed.

Let me know your thoughts

JohnGinnane:
Also this is still WIP. If we get some conclusion on what to do with other bots then I hope to have it ready for tomorrow or Thursday

code/game/machinery/bots/secbot.dm
```diff
@@ -26,10 +26,13 @@
 	radio_frequency = SEC_FREQ //Security channel
 	bot_type = SEC_BOT
 	model = "Securitron"
+	var/movement_delay = 4
+
+	#define BOT_MOVEMENT_DELAY 4 // This will be the default movement delay
```
Carbonhell: that's not a proper use of define. You can't make a define owned by something, and anyway why even use it here?You can access the value you assigned to vars from coding with initial(varname), so for example initial(movement_delay) will always return 4 in this case

code/game/machinery/bots/secbot.dm
```diff
@@ -326,6 +328,31 @@ Auto Patrol: []"},
...
+	var/obj/item/weapon/circuitboard/bot_upgrade_boost/B = locate(/obj/item/weapon/circuitboard/bot_upgrade_boost) in upgrades
+
+	if(B)
+		return !B.boost
+
+/obj/machinery/bot/secbot/proc/activate_boost()
+	var/obj/item/weapon/circuitboard/bot_upgrade_boost/B = locate(/obj/item/weapon/circuitboard/bot_upgrade_boost) in upgrades
+	
+	if(B)
+		B.boost = 1
```
Carbonhell: when you got a var that can be either 1 or 0, it'd be nicer to use TRUE and FALSE, just for the looks

JohnGinnane: Sounds good, that var was originally 0 (for off), 1 (for on), and 2 (for cooldown) but I can use true/false instead

code/game/machinery/bots/secbot.dm
```diff
@@ -326,6 +328,31 @@ Auto Patrol: []"},
...
+		movement_delay = B.boost_delay
+		
+		spawn(B.boost_length * 10)
+			deactivate_boost()
+
+/obj/machinery/bot/secbot/proc/deactivate_boost()
+	movement_delay = BOT_MOVEMENT_DELAY
```
Carbonhell: instead of setting this to a value wouldn't it be better to add the delay?so if the main movement delay value changes this will still be compatible

code/game/machinery/bots/secbot.dm
```diff
@@ -326,6 +328,31 @@ Auto Patrol: []"},
...
+		return !B.boost
+
+/obj/machinery/bot/secbot/proc/activate_boost()
+	var/obj/item/weapon/circuitboard/bot_upgrade_boost/B = locate(/obj/item/weapon/circuitboard/bot_upgrade_boost) in upgrades
+	
+	if(B)
+		B.boost = 1
+		movement_delay = B.boost_delay
+		
+		spawn(B.boost_length * 10)
```
Carbonhell: would be better to just make the var *10 when setting it,just add a comment to it saying that it's a value in deciseconds

code/game/machinery/bots/secbot.dm
JohnGinnane: Do you mean movement_delay = movement_delay + (initial(movement_delay) - B.boost_delay)?

code/game/machinery/bots/secbot.dm
JohnGinnane: Fair enough, the less magic numbers, the better!

code/modules/research/designs/machine_designs.dm
```diff
@@ -460,4 +460,10 @@
...
 	build_path = /obj/item/weapon/circuitboard/grounding_rod
-	category = list ("Misc. Machinery")
\ No newline at end of file
+	category = list ("Misc. Machinery")
+
+/datum/design/bot_upgrade_boost
+	name = "Bot Upgrade: Boost"
+	desc = "The circuit board for giving a speed boost to bots"
+	id = "bot_upgrade_boost"
+	req_tech = list("programming" = 2, "engineering" = 2)
```
Carbonhell: You missed the category and build type here,meaning that nothing can build it. I suggest making it be buildable from exosuit fabricators,since it's the better place.

code/game/machinery/computer/buildandrepair.dm
```diff
@@ -248,6 +248,13 @@
 	name = "circuit board (Library Visitor Console)"
 	build_path = /obj/machinery/computer/libraryconsole
 	origin_tech = "programming=1"
+/obj/item/weapon/circuitboard/bot_upgrade_boost
```
Carbonhell: Umm. By making it a circuitboard you may run into weird issues,such as this being able to be used to build a computer,which will cause weird shit. Maybe it'd be better to just define its own type under weapon/

JohnGinnane: It would still be possible to print it using the exosuit fabricator though, right?

Carbonhell:
Yea as i said it needs fixing, as for now it's not in a merge-able state

JohnGinnane:
Will be working on this tonight

code/game/machinery/bots/secbot.dm
JohnGinnane: This has been resolved

code/game/machinery/bots/secbot.dm
JohnGinnane: This has been implemented

code/game/machinery/bots/secbot.dm
JohnGinnane: Currently setting movement_delay to initial when the boost is over

code/game/machinery/bots/secbot.dm
JohnGinnane: This has been implemented

code/modules/research/designs/machine_designs.dm
JohnGinnane: I've moved all of this information into a new file as we talked about, and now it appears under the "Misc" section along with things like batteries

code/game/machinery/computer/buildandrepair.dm
JohnGinnane: This has been implemented

JohnGinnane:
Okay I've done lots more revisions and added the same boost to medibot, and added some fluff for when he explodes.

@Nebyab let me know what you think of this

code/game/machinery/bots/secbot.dm
```diff
@@ -26,10 +26,11 @@
 	radio_frequency = SEC_FREQ //Security channel
 	bot_type = SEC_BOT
 	model = "Securitron"
+	var/movement_delay = 4
```
Carbonhell: Why only this bot has this var?

code/game/machinery/bots/medbot.dm
```diff
@@ -363,6 +366,31 @@
...
+
+	if(B)
+		return !B.boost
+
+/obj/machinery/bot/medbot/proc/activate_boost()
+	var/obj/item/weapon/bot_upgrade/boost/B = locate(/obj/item/weapon/bot_upgrade/boost) in upgrades
+	
+	if(B)
+		B.boost = TRUE
+		speed = B.medbot_boost_delay
```
Carbonhell: where's this var/define/whatever declared(medbot_boost_delay)

code/modules/research/designs/bot_upgrades.dm
```diff
@@ -0,0 +1,43 @@
...
+/datum/design/bot_upgrade
+	// Should not be acquirable
+
+/datum/design/bot_upgrade/boost
+	name = "Bot Upgrade: Boost"
+	desc = "The circuit board for giving a speed boost to bots"
+	id = "bot_upgrade_boost"
+	req_tech = list("programming" = 2, "engineering" = 2)
+	build_type = MECHFAB
+	materials = list(MAT_GLASS = 500)
```
Carbonhell: isn't the cost a bit too low,i'd say atleast 1-2 metal and glass sheets

code/modules/research/designs/bot_upgrades.dm
```diff
@@ -0,0 +1,43 @@
...
+	build_path = /obj/item/weapon/bot_upgrade/boost
+	category = list("Misc")
+	construction_time = 100
+
+////////////////////////
+// CIRCUIT BOARD ITEM //
+////////////////////////
+
+/obj/item/weapon/bot_upgrade/
```
Carbonhell: ugly ending slash

code/modules/research/designs/bot_upgrades.dm
```diff
@@ -0,0 +1,43 @@
...
+	density = 0
```
Carbonhell, line 24: don't weapons have density & anchored set to 0 by default iunno

code/modules/research/designs/bot_upgrades.dm
```diff
@@ -0,0 +1,43 @@
...
+	anchored = 0
+	w_class = 2
+	name = "bot upgrade"
+	icon = 'icons/obj/module.dmi'
+	icon_state = "id_mod"
+	item_state = "electronic"
+	origin_tech = "programming=2"
+	materials = list(MAT_GLASS=500)
+	var/board_type = "computer"
```
Carbonhell, line 33: ???

code/game/machinery/bots/secbot.dm
JohnGinnane: This was originally a magic number further down the code. I'm thinking of refactoring all the bots so they share the same movement variables

code/modules/research/designs/bot_upgrades.dm
JohnGinnane, line 24: A lot of this file was copy and pasted, and I'm not sure what exactly some of these properties do

code/modules/research/designs/bot_upgrades.dm
JohnGinnane: Is there a comparable item whose value I could use, as I'm unsure how much metal exactly 1 sheet is

code/modules/research/designs/bot_upgrades.dm
danhibiki, line 24: By default obj/item/weapon has no density set, but its probably safer and not a huge issue to define density as zero.

code/modules/research/designs/bot_upgrades.dm
danhibiki: One sheet should be 2000 units for either material, metal or otherwise.

JohnGinnane: Okay, I'll change to 4,000 metal, 2,000 glass

code/modules/research/designs/bot_upgrades.dm
JohnGinnane, line 33: More copy paste (read: bad habit) I believe this variable is just verification for when you're building computers, so this actually needs to be removed

JohnGinnane:
I've made changes to make the movement a little more consistent and also moved the boost related procs into the parent type bot because they now behave the same for medbot and secbot

myontasm:
Is there any way we can make bots like these turned into proper mobs instead of objects?

Lawligagger:
Pretty sure there's a way to

JohnGinnane:
Would simple animal be enough?

Carbonhell:
They should be simple animals, but iunno if there's anyone up for doing it as for now. And anyway the PR is about a different subject so iunno, if this gets merged porting this won't be hard at all

Carbonhell:
So?Any news?

JohnGinnane:
Haven't had a chance to look at this. Hope to this weekend where I've more free time. Do you want to leave the change to mob until later? Would make it easier to track changes and it can all be done in one PR

JohnGinnane:
Just waiting on approval as the move from object to mob should be another pull request

myontasm:
Yeah definitely, I was just asking if we could do that in the future, not necessarily now.

As for this, I'll give it a +1 for now.

KayeArray:
beepsky can already easily shut down meth heads this shit is not needed

Carbonhell:
Object to mob has to be in another PR since it'll be an huge change which might need to be reverted if it fucks up badly. It might also be a bit too hard for you, it involves controllers. But trying won't hurt.Code's okay for now by the way

Carbonhell:
Oh crap,won't this conflict with crystal frightbot&thatotherbot pr?

Carbonhell:
Actually i'm not sure if code is 100% ok, i don't have much time to check today, but at a fast look it looked fine

JohnGinnane:
You're right, this might conflict with Crystal's PR. In that case let's wait until his is accepted and then I can pull down latest version and resolve any conflicts

Nokington:
:+1:

Carbonhell:
No merge conflicts, can it be merged @JohnGinnane ?

code/game/machinery/bots/bots.dm
```diff
@@ -273,6 +276,21 @@
 				..()
 				healthcheck()
 
+	if(istype(W, /obj/item/weapon/bot_upgrade/boost))
+		if(!open)
+			user << "<span class='warning'>You cannot upgrade [src] with the maintenance panel closed!</span>"
+			return
+		else if(locate(/obj/item/weapon/bot_upgrade/boost) in upgrades)
+			user << "<span class='warning'>[src] already has that upgrade installed!</span>"
```
Carbonhell, line 284: forgot the return seems

code/game/machinery/bots/medbot.dm
```diff
@@ -309,7 +310,7 @@
 			stunned = 0
 		return
 
-	if(frustration > 8)
+	if(frustration >= 8)
```
Carbonhell, line 313: why this =?

code/game/machinery/bots/medbot.dm
```diff
@@ -346,12 +347,19 @@
...
+		walk_to(src, patient, 1, movement_delay)
+
+		if((get_dist(src, patient)) >= (olddist))
+			frustration++
+		else
+			frustration = 0
 		return
 
-	if(path.len > 8 && patient)
+	if(path.len >= 8 && patient)
```
Carbonhell, line 362: why =?

code/game/machinery/bots/medbot.dm
JohnGinnane, line 313: To be consistent with https://github.com/HippieStationCode/HippieStation13/blob/master/code/game/machinery/bots/secbot.dm#L211

Crystalwarrior, line 313: more or equeal to????

code/game/machinery/bots/medbot.dm
Crystalwarrior, line 362: same as above??

Carbonhell:
Code seems ok,just explain the two =s you've put

code/game/machinery/bots/medbot.dm
JohnGinnane, line 313: It would result in it basically being > 7. It's completely the same. I just changed the medbot to have the same range as the secbot. If you want I can revert, but this makes the medbot more determined by 1 tile

code/game/machinery/bots/medbot.dm
JohnGinnane, line 362: See above

JohnGinnane:
See https://github.com/HippieStationCode/HippieStation13/pull/2078#discussion_r59132254

Carbonhell:
Code ok and 3 heads ok,merging

PR #2089, opened 2016-03-31
Adds the ability to name guardians on manifestation as a guardian user

ghost:
General code for changing names work, although if anything needs to be tacked on before it can be merged just tell me.

### DIFF
--- a/code/modules/mob/living/simple_animal/guardian/guardian.dm
+++ b/code/modules/mob/living/simple_animal/guardian/guardian.dm
@@ -600,6 +600,8 @@
 				if(world.time < time_limit)
 					if(length(guardianNewName) > 0)
 						G.name = guardianNewName
+					else
+						user << "<span class='danger'>Sorry, you can't have a null name!</span>"
 				else
 					user << "<span class='danger'>Sorry, you didn't give a new name in time!</span>"
 	else
--- a/code/modules/mob/living/simple_animal/guardian/guardian.dm
+++ b/code/modules/mob/living/simple_animal/guardian/guardian.dm
@@ -593,6 +593,15 @@
 	if(candidates.len)
 		theghost = pick(candidates)
 		spawn_guardian(user, theghost.key)
+		for(var/mob/living/simple_animal/hostile/guardian/G in living_mob_list)
+			if (G.summoner == user)
+				var/time_limit = world.time + 300
+				var/guardianNewName = stripped_input(G.summoner, "You are the user of [G.name]. Would you like to name your guardian something else?", "Name Guardian", G.name)
+				if(world.time < time_limit)
+					if(length(guardianNewName) > 0)
+						G.name = guardianNewName
+				else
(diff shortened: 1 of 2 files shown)


JohnGinnane:
Please note we tested this and it's working correctly

myontasm:
:+1: star pratinum

code/modules/mob/living/simple_animal/guardian/guardian.dm
```diff
@@ -593,6 +593,17 @@
 	if(candidates.len)
 		theghost = pick(candidates)
 		spawn_guardian(user, theghost.key)
+		for(var/mob/living/simple_animal/hostile/guardian/G in living_mob_list)
+			if (G.summoner == user)
+				var/time_limit = world.time + 300
+				var/guardianNewName = stripped_input(G.summoner, "You are the user of [G.name]. Would you like to name your guardian something else?", "Name Guardian", G.name)
```
Carbonhell: You've put no limit in this meaning that the name's limit is,by default,MAX_MESSAGE_LEN. You need to add an argument, which i suggest to be MAX_NAME_LEN  
cause from what you can see here:
#define MAX_MESSAGE_LEN         1024
#define MAX_NAME_LEN            26
#define MAX_BROADCAST_LEN       512
having the name be of 1024 characters would be weird, but 26 chars look perfect

code/modules/mob/living/simple_animal/guardian/guardian.dm
```diff
@@ -593,6 +593,17 @@
 	if(candidates.len)
 		theghost = pick(candidates)
 		spawn_guardian(user, theghost.key)
+		for(var/mob/living/simple_animal/hostile/guardian/G in living_mob_list)
```
Carbonhell: why this loop?can't you just find the mob based on the key you have,somehow?

ghost: I think the guardian only has a reference to the summoner and not the other way around

The part above for checking to see the summoner already does the loop thing. I would imagine if there was a direct reference then it would do a check to see if the user has one.

Carbonhell: https://github.com/HippieStationCode/HippieStation13/blob/7fb5bbbfcc6e1b6880f70c837ea9617df509b4d8/code/modules/mob/living/simple_animal/guardian/guardian.dm#L659
Just add a "return G" so this proc will return the guardian, then you edit the line up here(   spawn_guardian(user, theghost.key)   ) to var/mob/living/simple_animal/hostile/guardian/G = spawn_guardian(user, theghost.key)

Carbonhell:
code is ok

ghost:
If code is gucci then it's ready for merging

Carbonhell:
it's a feature,it needs a second head approval

Nokington:
k

Pyko1:
2 aprovals and apparently code is fine, merging

PR #2090, opened 2016-03-31
Greater Mindslave Implants

Sarcalogo:
Adds Greater Mindslave Implants. 

They're like Mindslave Implants except they last forever. Cost 16 TC. Don't work on Traitors or Changelings.

I also fixed up the messages you receive when being mindslaved. They were NOT very good, and looked bad when you were actually implanted.

### DIFF
--- a/code/game/objects/items/weapons/implants/implant_mindslave.dm
+++ b/code/game/objects/items/weapons/implants/implant_mindslave.dm
@@ -1,4 +1,4 @@
-/obj/item/weapon/implant/mindslave
+/obj/item/weapon/implant/mindslave/
 	name = "mindslave implant"
 	desc = "Now YOU too can have your very own mindslave! Pop this implant into anybody and they'll obey any command you give for around 15 to 20 minutes."
 	origin_tech = "materials=2;biotech=4;programming=4"
@@ -29,8 +29,7 @@
 			return 0
 		if(..())
 			target << "<span class='notice'>You feel a surge of loyalty towards [user].</span>"
-			target << "<span class='userdanger'> You MUST obey any command given to you by your master(that doesn't violate any rules). You are an antag while mindslaved.</span>"
-			target << "<span class='danger'>You CANNOT harm your master. Check your memory ( with the notes verb) if you forget who your master is.</span>"
+			target << "<span class='danger'>You have been mindslaved by [user]! Until the implant expires, they are your master. You MUST obey any command given to you and CANNOT harm your master unless they order you to do so.</span>"
 			var/time = 9000 + rand(60,3000)
 			timerid = addtimer(src,"remove_mindslave",time)
 			if(!target.mind.special_role)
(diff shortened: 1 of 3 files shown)


code/game/objects/items/weapons/implants/implant_mindslave.dm
```diff
@@ -1,4 +1,4 @@
-/obj/item/weapon/implant/mindslave
+/obj/item/weapon/implant/mindslave/
```
Carbonhell, line 1: Why this ending slash

code/game/objects/items/weapons/implants/implant_mindslave.dm
```diff
@@ -72,4 +71,62 @@
 /obj/item/weapon/implanter/mindslave/New()
 	imp = new /obj/item/weapon/implant/mindslave(src)
 	..()
+	update_icon()
+
+/obj/item/weapon/implant/gmindslave
```
Carbonhell, line 76: Ugly copypasta, i suggest grouping those in 1 item

Sarcalogo:
I don't know what you're talking about.

Carbonhell:
Oh,i though you were zadeon,woops, well the code isn't really perfect and it needs improvement,unsure if you can work on it or not?

Sarcalogo:
Okay... what are the issues? I was going to make it a child of mindslave implant, but when I did that I got TWO mind slave messages and some other silly little things.

Carbonhell:
You need to make it a child of mindslave implants but you also need to make a var that handles time, and set it to the normal time for normal mindslave implants and,for example, -1 for those infinite mindslave implants. Iunno what you did to get two mindslave messages,i'd have to check what you did to correct it

Sarcalogo:
I'll be clear, I have no idea how BYOND coding works. I don't know how to override the parent functions or methods or whatever the hell it is called that causes that issue. I guess I can try again now that everything is ironed out and it works perfectly to see if I can rewrite it that way.

Carbonhell:
just do /obj/item/yourprocname(arguments) then write code,that's how you override the parent function. if you want to call the parent function you do ..(), through this may be a bit hard for you to do if you don't know this stuff

Sarcalogo:
Okay. The issue is that I'm trying to avoid copy pasted code correct? What I presume you want me to do is to inherit the methods from the super class, mindslave, and only then add in the lines of code that are different? So instead of having this very long

/obj/item/weapon/implant/mindslave/gmindslave/implant(mob/target,mob/user)

It should be much shorter, and only include:

```
if(target.mind in ticker.mode.traitors)
...
return 0
```

Is that is what is supposed to happen? Or is that if there is any difference between the functions, or procs, or whatever they are called, I have to rewrite the whole thing, even if for the most part they are the same? If that is the case, there is nothing to inherit, because all of the procs or whatever are different

Sarcalogo:
So maybe then something like this?

```
/obj/item/weapon/implant/mindslave/gmindslave/implant(mob/target,mob/user)
..()
if(target.mind in ticker.mode.traitors)
...
return 0
```

Carbonhell:
it'd be better to name it "permanent" rather than "gmindslave", firstly. then, you need to edit a lot of stuff,editing one proc won't be enough. then you need to decide when to call ..() depending on what it does,i think you may want to call it at the end

Sarcalogo:
Okay I think I've figured it out for the most part. Let me try again.

code/game/objects/items/weapons/implants/implant_mindslave.dm
Crystalwarrior, line 1: Consistency i'd imagine?
EDIT: hold on that's the opposite of consistency

Sarcalogo:
Yeah... code is not working as intended... this is what I've tried and it compiles, but it functions the exact same way as a normal mindslave implant does. 

```
/obj/item/weapon/implant/mindslave/gmindslave/implant(mob/target,mob/user)
	if(target.mind)
		if(..())
			if(target.mind in ticker.mode.traitors)
				target.visible_message("[target] seems to resist the implant!", "You feel a familiar sensation in your head that quickly dissipates.")
				return 0
			if(target.mind in ticker.mode.changelings)
				target.visible_message("[target] seems to resist the implant!", "You feel a familiar sensation in your head that quickly dissipates.")
				return 0
			if(..())
				target << "<span class='userdanger'>You have been mindslaved by [user]! From now on, they are your master. You MUST obey any command given to you and CANNOT harm your master unless they order you to do so.</span>"
				if(!target.mind.special_role)
					target.mind.special_role = "Permanent Mindslave"
					ticker.mode.traitors |= target.mind
	..()		
	
	return 0
```

Sarcalogo:
Hmm, maybe this is the issue...

Does this fix it?

`if(..(..()))`

Carbonhell:
no, ..() is like calling the proc with the same name of the parent so doing ..(..()) cannot work,what are you even trying to do,i'm heavily confused, maybe you'd want to join our slack chat iunno

Sarcalogo:
Right, the parent of 
/obj/item/weapon/implant/mindslave 
is implant.

The parent of /obj/item/weapon/implant/mindslave/gmindslave 
is mindslave

In the mindslave implant, the original one, this piece of code exists.
```

if(..())
		
```
I just copy pasted that into the greater mindslave, which means that if statement now refers to mindslave instead of implant

Sarcalogo:
Basically I just want to call a function from the parent of the parent instead of the parent.

Carbonhell:
Put a ..() in the parent proc then

Sarcalogo:
That's what I did.
if(..(..()))

Carbonhell:
No, doing ..() is like calling a proc named "..", which means the parent type. Inside () you can only put arguments like mob/user(but most of the time there's no reason to do so unless you changed an arg midway in your proc). You can't put a ..() inside a ..()

Sarcalogo:
Yeah dude honestly, none of the shit I'm trying works after making it a child to mindslave.

I tried calling ..() at the beginning, It didn't work.
I tried calling ..() at the end, it didn't work.
I tried not calling..() at all, but now despite all the code being the same as the functional implant I pull requested here, the Greater MindSlave implant functions exactly the same as the lesser one.

Nebyab:
I mean if it ends up working, sure alright.

Sarcalogo:
It works now, it's just not up to snuff or something. I'll try again to get it functioning while as a child to mind slave implant, but I don't think I will be able to.

Limed00d:
Honestly feel that this is way better than the temporary mindslave implants as they are in my opinion purdy shite. So, yeah sure :+1:

Lawligagger:
I'm down

PR #2102, opened 2016-04-02
Debug message for identity bans

Carbonhell:
adds a debug message to see if this fucking identity ban list is being actually loaded or not,must be called through advanced proccall
there's also a possible fix but i doubt that will fix it


Crystalwarrior:
Seems alright, merging

PR #2104, opened 2016-04-02
Wrestling Belt traitor item, CQC gloves buff

Crystalwarrior:
Changes: 
-  - tweak: "Made CQC gloves have 100% clinch chance"
-  - tweak: "Removed stamina damage applied to attacker for Wrestling Belt"
-  - rscadd: "Wrestling Belt is now an uplink item! Costs 12 TC, same as CQC gloves"
-  - bugfix: "Fixed runtime error with Wrestling Belt" 

CQC gloves are still hard to use especially on aware opponents. 100% clinch chance makes surprise attacks not depend on awkward directions and RNG but instead relies on your speed alone.
Wrestling belt is added as a traitor weapon as well for the same price of 12 TC.
Wrestling Belt received a buff since initially it was designed for civvie use but now it can be very viable in 1v1 situations. Stamina drain for some of the special moves was removed.

### DIFF
--- a/code/datums/uplink_item.dm
+++ b/code/datums/uplink_item.dm
@@ -459,6 +459,12 @@ var/list/uplink_items = list()
 	item = /obj/item/clothing/gloves/cqc
 	cost = 12
 
+/datum/uplink_item/dangerous/wrestling
+	name = "Wrestling Belt"
+	desc = "A powerful wrestling moveset given to you via championship belt."
+	item = /obj/item/weapon/storage/belt/champion/wrestling
+	cost = 12
+
 /datum/uplink_item/dangerous/reinforcement
 	name = "Reinforcements"
 	desc = "Call in an additional team member. They come with predefined gear sets, but you might want to save some telecrystals to arm them better."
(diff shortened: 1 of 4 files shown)


Carbonhell:
code's ok

Nebyab:
ok

JohnGinnane:
What if we made CQC and belt both cost 10? That way someone could roleplay as a wrestler and just run around the station using their fighting skills to murderbone

Crystalwarrior:
I'd rather have traitors make a choice between either wrestling or CQC since you can't really combine both either way.

JohnGinnane:
Ah okay

Kierany9:
How powerful is the wrestling belt now exactly? Keep in mind they're available to Sleeping Carp gangsters for 20 Influence and that they have a martial arts scroll for 30 Influence.

Crystalwarrior:
Still not that powerful. Wrestling Belt is great in 1v1 fights but if it's a crowd fight you gotta have some support.
Sleeping Carp gangsters will always go for the sleeping carp itself anyway due to that instakill crit move.

myontasm:
Sure.

Lawligagger:
https://www.youtube.com/watch?v=CysDaXy44Us

yes

PR #2106, opened 2016-04-02
Re-nerfs superfarts

Kokojos:
Because superfarts were never re-nerfed and still do 75.

I don't remember the orignal ammount they were nerfed to, 40 seems fine to me.

### DIFF
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -530,8 +530,8 @@
 						for(var/mob/living/M in range(0))
 							if(M != src)
 								visible_message("\red <b>[src]</b>'s ass blasts <b>[M]</b> in the face!", "\red You ass blast <b>[M]</b>!")
-								M.apply_damage(40,"brute","head")
-								add_logs(src, M, "superfarted on", object=null, addition=" (DAMAGE DEALT: 40)")
+								M.apply_damage(50,"brute","head")
+								add_logs(src, M, "superfarted on", object=null, addition=" (DAMAGE DEALT: 50)")
 
 						visible_message("\red <b>[src]</b> blows their ass off!", "\red Holy shit, your butt flies off in an arc!")
 
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -530,8 +530,8 @@
 						for(var/mob/living/M in range(0))
 							if(M != src)
 								visible_message("\red <b>[src]</b>'s ass blasts <b>[M]</b> in the face!", "\red You ass blast <b>[M]</b>!")
-								M.apply_damage(75,"brute","head")
-								add_logs(src, M, "superfarted on", object=null, addition=" (DAMAGE DEALT: 75)")
+								M.apply_damage(40,"brute","head")
+								add_logs(src, M, "superfarted on", object=null, addition=" (DAMAGE DEALT: 40)")
 
 						visible_message("\red <b>[src]</b> blows their ass off!", "\red Holy shit, your butt flies off in an arc!")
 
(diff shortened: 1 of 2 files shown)


Kierany9:
They were nerfed to 50 IIRC

Carbonhell:
--------------------------------------1
no really atleast 50 please

Crystalwarrior:
:+1: :+1: :100: 
Superfarts were made a little too fucking powerful. 75 damage is DEVASTATING and from what I can tell it ignores armor too.
Wait why the fuck does it ignore armor!???

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -530,8 +530,8 @@
 						for(var/mob/living/M in range(0))
 							if(M != src)
 								visible_message("\red <b>[src]</b>'s ass blasts <b>[M]</b> in the face!", "\red You ass blast <b>[M]</b>!")
-								M.apply_damage(75,"brute","head")
-								add_logs(src, M, "superfarted on", object=null, addition=" (DAMAGE DEALT: 75)")
+								M.apply_damage(40,"brute","head")
```
JohnGinnane: Could we apply this to a random body part instead?

Crystalwarrior: I think it's better to stay this way so you can wear head protection to not be utterly fucking destroyed (ASSUMING @Kokojos IMPLEMENTS ARMOR CHECKS)

JohnGinnane: What if we checked whether the target and src are on the ground? it seems odd if someone is crit lying on the ground and they're able to superfart on the head of someone who's standing up

Crystalwarrior: chances are people will throw a shitfit because "muh clutch victory", and also since superfart is the only choice people have left if they've been downed by their opponent.

Crystalwarrior: Not to mention superfart is only giving them one shot to retaliate.

Pyko1:
there should be quite a lot damage done to the person who has his ass blown off...

Crystalwarrior:
75 damage is overkill since it's near-instant crit (/tg/ bleeding made it MUCH worse).

ghost:
this was an obvious fuel by him getting fucekd like 3 rounds ago
biased

code/modules/mob/living/carbon/human/emote.dm
JohnGinnane: Good points

JohnGinnane:
What if superfart started at lower damage (e.g. 25 or 40) and then scaled with how fat (or how much stuff they've consumed) they are up to a max of 75?

Crystalwarrior:
That's a little too finnicky lol

JohnGinnane:
You could use the hungry variable, which gets set to like 0 when they superfart anyway

Crystalwarrior:
Anyway I think the superfart should dish out 50 damage OR have armor checks. 75 unobstructed damage is overkill when the risk of using superfart is so incredibly low - slow movespeed doesn't matter if you're already downed by your opponent and about to be killed.

Nebyab:
50 without armor checks and i'll okay this

JohnGinnane:
How does base 25, max 75 sound, all damage takes armour into account? The damage based on hungry level, and maybe half this damage is applied to the owner.

Nokington:
what neb said

Lawligagger:
http://i.imgur.com/ygJ7Oud.png

Carbonhell:
fix it kokojo FIX IT

Kokojos:
Changed to 50, assuming it's pre-okayed by 3 heads

Crystalwarrior:
Aye assuming this is a-ok to merge

blobbernaut:
how the fuck did this get un-nerfed???

PR #2107, opened 2016-04-02
Reverts april fools shit

Crystalwarrior:
Title, a basic revert of april fools.


Carbonhell:
-1,we should keep lobby screen and title soundfiles since it's april now, plus i really liked some sounds for example the "NO" when you get hit by a gun,which i strongly believe we should keep. They add a kickass atmosphere imo

Crystalwarrior:
are you seriously intending to keep the sounfiles and the lobby screen for the entirety of the goddamn month
because if you are you are absolutely out of your mind, I won't be able to deal with the earrape for a whole fucking month.
The "kickass atmosphere" already got old fast. We should implement some of these sounds as toys or something instead.

TDcoolguy300:
I'm just going to merge it

Nebyab:
Its a revert. None of this was intended to kept. We can bring somethings back after the fact, but we should put things back to business as usual.

PR #2110, opened 2016-04-03
Fixed washing machine not washing hoods and other contents

### LINKED ISSUE #1743: Cleaning an Engineering Hardsuit (or any hardsuit with attached helmet probably) does not clean the helmet.
If you put an engineering hardsuit covered in blood in a sink, it will clean the body just fine but the helmet will stay dirty. Trying to clean blood off fully is impossible unless you have soap in this case.

### LINKED ISSUE #2083: Coloring a winter coat with a crayon in the washer doesn't give the hood the same color as the coat.
TBH it just looks silly. The hood is colored in one sprite, but not colored on the person or the head slot.

image

JohnGinnane:
The washing machine will now wash all the contents within the items that are being washed, as well as colour them

Fixes #1743 and fixes #2083


code/game/machinery/washing_machine.dm
```diff
@@ -64,14 +69,23 @@
...
 						break
+
+				for(var/obj/item/clothing/J in I.contents)
+					J.color = wash_color
+					J.desc = J.desc + (length(J.desc) > 0 ? " " : "") + "It looks soaked in [wash_color] tincture."
+
+					for(var/old_color in colors)
+						if(findtext(J.name, old_color))
+							J.name = n_replace(J.name, old_color, wash_color)
+							break
```
Crystalwarrior, line 88: I dunno, maybe there's  better way to do this than copy-pasting code?
Perhaps, you could have double for-loops checking contents of both src and I, add all found objects to the var/list/L and then for(var/obj/item/clothing/I in L you'd execute the code?
Though that would mean THREE for loops as opposed to two. Fuck.

JohnGinnane, line 88: I thought about using recursion to search all contents but I figure a depth of 2 should be plenty

Crystalwarrior:
Changelog included, this is a fix so merging.

PR #2111, opened 2016-04-03
DELETE THIS *DON'T MERGE*

Zadeon:
This should fix the kamikazes randomly blowing up. It was just checking for any mob and not mob/living.

 Also when I compiled this it gave me a warning about the sound file so that's why I re-added it just to be sure.


Zadeon:
Oh crap what did I do wrong.

Please delete this actually give me a second to get my real branch.

PR #2115, opened 2016-04-03
Made dental implant surgery accept only pills

### LINKED ISSUE #2080: Dental implants accept chem patches due to them being child of pills
HOW

JohnGinnane:
It now checks if the tool is a pill type and the parent_type is not pill type, to disable patches from being used

Fixes #2080


Crystalwarrior:
Code seems A-OK :+1:

code/modules/surgery/dental_surgeries.dm
```diff
@@ -77,7 +77,12 @@
 	time = 16
 
 /datum/surgery_step/insert_pill/preop(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
-	user.visible_message("[user] begins to wedge \the [tool] in [target]'s [parse_zone(target_zone)].", "<span class='notice'>You begin to wedge [tool] in [target]'s [parse_zone(target_zone)]...</span>")
+	if(istype(tool, /obj/item/weapon/reagent_containers/pill) && tool.parent_type != /obj/item/weapon/reagent_containers/pill)
```
Carbonhell: Instead of this,just do a strict type check, which means excluding childs. if(type == /obj/item/weapon/reagent_containers/pill)

JohnGinnane: Ah, makes much more sense!

meteoses:
Thanks, as I said in the previous thread: I approve of a direct bug fix rather than trying to use this to springboard more nerfs.

JohnGinnane:
Made the condition make more sense

myontasm:
Yeah sure :+1:

Lawligagger:
a ok

PR #2130, opened 2016-04-07
Fixed stuns being 100% blocked by any energy armor value

Crystalwarrior:
fixed an incredibly strange and stupid bug causing stuns to be 100% useless against armor with ANY anti-stun value.

Also fixed electrodes exploding in sparks while hitting armor unless 100% blocked (pre-crawl update issue that was not as noticeable)

also add "tg fucked up" label pls :^)


Carbonhell:
Code's ok,merging when travis passes

carlarc:
well not REALLY, I could get knocked down by a telebaton
but no one is going to see this

PR #2138, opened 2016-04-08
Cleans up changelog directory and updates core changelog.

nalarmg:
Note - the changelog gets a little confused when I do this. Presumably the date that appears is the date the changelog is actually updated, not when the update was done, which is a bit silly.


nalarmg:
DNM whilst I attempt to fix the changes all appearing under one date.

PR #2141, opened 2016-04-09
Updates Templates, fixes the powernet issues.

TDcoolguy300:
Forces templates to regenerate powernets when you spawn shit in. Updated the maps to include the new chair/stool/bed paths. Fixed the alien abductor gun in the Alien ship


TDcoolguy300:
Fucking, looks like i fucked up with something, give me a bloody moment

TDcoolguy300:
Looks like i commited backups i accidentally made. Should be good to go

Carbonhell:
why two makepowernets?when is template initialize proc called?shouldn't it be called only when placing a template

PR #2142, opened 2016-04-09
Stun Update 2.0

Crystalwarrior:
Alright, so last time this got in a little WAY too fast without any semblance of discussion and, naturally, the server was in shambles.
Let's try this again, but this time we need to discuss all the balance changes applied in this PR, perform extensive testing and figure out if I missed any stuns that give you too much "crawl space".

//**DISCLAIMER**//
Most affected are stuns that either cannot be followed up with a reliable cuff or are long enough on their own to warrant control given back to the victim so they're not fucked for an extremely long time (some 20-22 second stuns, for example). Some of "environmental" stuns were also affected.
//**DISCLAIMER END**//

Keep in mind that due to the tick nature of Weaken/Stun procs, the "first tick" will have a slightly random effect on stun duration - basically depends on how early into the tick the proc was called.

Example: Stun(2) might last from 2 to 4 seconds due to 1 tick taking 2 seconds and the ticks aren't initialized in the Stun() proc itself.

**STUNS THAT WEREN'T TOUCHED/YOU CANNOT CRAWL FROM/CAN ONLY CRAWL FROM ON LAST TICK:**
- Lichdom spell stun for wizard
- Being absorbed by changelings (after being stabbed with a proboscics)
- Stun trap in Hand of God
- Most shadowling abilities
- Defib/emagged Defib stuns
- Flashbangs and Flashes (can crawl only on last tick)
- Stunbatons and peppersprays
- CQC gloves and Krav Maga, as well as Plasma Fist's knockdown punch
- Xeno Hunter Pounce
- Fire resist (rolling on ground to put yourself out)
- Xenomorph tackles
- Plant people having their vacuoles boiled from floramut
- Being exhausted from low stamina
- Changelings being stunned from when they lose their horror form due to damage
- Venus Plant human trap
- Untying the noose over your neck when hanged
- Detective revolver
- Neurotoxin spit on xenos
- Ebow
- Most reagents were untouched (other than flash powder, stun duration was set to be the same as flashes)
- Butt smacking you in the face
- Wizard's magic missile
- All pool-induced stuns
- Tabling
- Slam Dunking someone in holo-hoop (4 stun, 5 weaken)
- Electropack zap (4 stun, 5 weaken)
- Walking on shards/dice/etc. w/o shoes (3 ticks stun, 5 ticks weaken)
- Strained Muscles stun for changelings
- Airlock charge disarm fail stun
- Reagent explosion
- Spraycan spray
- Tray's 10% stun chance
- Wallslam stun
[...]

PR #2146, opened 2016-04-09
Made space lube able to be chained

Crystalwarrior:
Lube can now be chained. That means if you have a hallway of lube people will keep going and going until the lube ends or until they've travelled 64 tiles.

Note that if https://github.com/HippieStationCode/HippieStation13/pull/2142 gets merged it also means lube can be crawled through safely, but you'll still be incredibly slow - slower than walking in fact.

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -282,17 +282,27 @@
 		C.Stun(s_amount)
 		C.Weaken(w_amount)
 		C.stop_pulling()
+		if(C.lying != oldlying) //did we actually fall?
+			var/dam_zone = pick("chest", "l_hand", "r_hand", "l_leg", "r_leg")
+			C.apply_damage(2, BRUTE, dam_zone)
+
 		if(buckled_obj)
 			buckled_obj.unbuckle_mob()
 			step(buckled_obj, olddir)
 		else if(lube&SLIDE)
-			for(var/i=1, i<5, i++)
-				spawn (i)
-					step(C, olddir)
+			var/i = 1
+			var/l = 0
+			while(i < 5 && l < 64) //Only 64 loops are allowed max to prevent infinite looping
+				l++
+				if(l%5==0)
 					C.spin(1,1)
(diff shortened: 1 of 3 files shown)


carlarc:
yes

AliiTea:
This is fucking great

JohnGinnane:
Code checks out and Crystal, Kierany, and I all tested this on a server and it works perfectly!

Spacedong:
glorious

Gandalf2k15:
This is fuckin great.

Nokington:
k

ghost:
yes please

McHatters:
Before we do this can we update the rules regarding lube

carlarc:
i see no controversy over here

Nokington:
Agreed. DNM until we actually make it clear that lubing makes you valid.

Kierany9:
It's glorious.

Carbonhell:
code and idea are both a+

Crystalwarrior:
Dunno why the DNM tag, isn't the lube situational rules-wise? Like, if you use it to kill people then you get valid but if you just use it in self defence it should be okay.

Nokington:
It was like 2am when I dnm. It doesn't really need it but I would prefer to keep it until we make it clear to players that luring makes them valid

Nebyab:
Lubing in the hallway makes you valid because you're being a cunt. Moreso with chain lube as a thing. We don't need to specifically enumerate everything. 

I'll approve it, but I have concerns that it might get irritating a fuckabout-y.

carlarc:
Make it a admin notice (_HAHA_) and make an announcement every 3 hours for 4 days.

meteoses:
This was on NoX back when I played it, loved it, you could chain slip someone gloriously into space.

myontasm:
Sounds great, but I'd rather wait until the "lubing makes you valid" rule is put up in black and white.

Lawligagger:
gonna merge

PR #2159, opened 2016-04-13
Increases Plasma Fist Cost to 2

TheOneAndOnlyCreeperJoe:
Because right now a meager cost of 1 is way too cheap for what it brings to the table in comparison to other combat spells.
- It can be used by robeless wizards
- It has the Disintegrate ability as well as several other abilities added to it for free.
- The combos are incredibly quick to execute if you are using hotkeys. In fact, you can gib two people with a single magic missile volley before they even can get up provided they both were stunned at the same time and are in close proximity of eachother.
- None of the combos have a cooldown, which allows you to chain Plasma Fists and basically gib a lot of people in short succession

The only thing Disintegrate has going for it over Plasma Fist is the ability to insta-kill and it's ability to function on non-humans (cyborgs for example), which is already substantial. But right now, Plasma Fist is just in too good of a spot for what it brings to the table in comparison to the rest, upping the cost is a step towards balancing it out.

(Removed the cost var entirely for Plasma Fist since it defaults to 2)

### DIFF
--- a/code/game/gamemodes/wizard/spellbook.dm
+++ b/code/game/gamemodes/wizard/spellbook.dm
@@ -347,7 +347,6 @@
 	desc = "Consider this more of a \"Spell Bundle\". This artifact is NOT reccomended for weaklings. An ancient scroll that will teach you the art of Plasma Fist. With it's various combos you can knock people down in the area around you, light them on fire and finally perform the PLASMA FIST that will gib your target."
 	item_path = /obj/item/weapon/plasma_fist_scroll
 	log_name = "PF"
-	cost = 1
 
 /datum/spellbook_entry/item/bloodbottle
 	name = "Bottle of Blood"
--- a/html/changelogs/Creeper Joe - Wizard Balance.yml
+++ b/html/changelogs/Creeper Joe - Wizard Balance.yml
@@ -4,4 +4,3 @@ delete-after: True
 
 changes:
   - tweak: "The Plasma Fist scroll now costs 2 instead of 1."
-  - tweak: "The Wizard's Lightning Bolt charges twice as fast."
--- a/html/changelogs/Creeper Joe - Wizard Balance.yml
+++ b/html/changelogs/Creeper Joe - Wizard Balance.yml
@@ -1,0 +1,7 @@
+author: Creeper Joe
+
+delete-after: True
+
+changes:
+  - tweak: "The Plasma Fist scroll now costs 2 instead of 1."
+  - tweak: "The Wizard's Lightning Bolt charges twice as fast."


Crystalwarrior:
Hah, when I first ported Plasma Fist I was pressured into making the cost 1 because it was too hard to use or soemthing. :+1: from me

trifornt-tinfornt:
Honestly I disagree with this. In order to make Plasma Fist that great you need to combine it with Mutate, and while it is true if you're really skilled you can pull off several disintegrates it's not like the standard disintegrate where all you have to do is walk up and punch someone. Otherwise it's honestly kind of a downgrade, requiring five punches to actually gib someone and unless you're fast with the thing then it won't be as effective as you're making it out to be.

TheOneAndOnlyCreeperJoe:
It is not effective without a stun.
It is incredible with a stun.

You hit incredibly fast anyway with plasma fist and if youre good with hotkeys you can gib someone in 1.5 seconds, which is enough time to plasma fist someone affected by any stun.

Disintegrate was most of the time used on targets stunned by magic missile anyway. If you are yakkety saxing after someone with disintegrate then you are doing it wrong.

That and a robeless disintegrate is cancer since youll likely have looted the armory anyway

trifornt-tinfornt:
Yeah, robeless is caner and I'll agree with that. Still don't really think increasing its price is the right way to go - seeing the past of the Power Fist which used to cost 14 TC, until it got bumped down to 10 TC, and then got nerfed and bumped up in price because people got good with it and then it got bumped down to 10 TC and nobody uses it any more I fear the same might happen here.

Crystalwarrior:
Thinking about it, it should probably stay at its current price point. Plasma Fist is not only a little hard to use, but it's also quite a spectacle. Wizards using Plasma Fist are incredibly more entertaining than wizards using disintegrate.
I always considered it to be a little gimmicky IMO

KayeArray:
No real need to increase the cost, it only works effectively with mutate, executing the combo takes a couple of seconds unless you have perfect timing and entering close combat even with mutate is risky as fuck.

Crystalwarrior:
Yeah, disintegrate is one-click-win making it viable in crowd fighting.
Plasma Fist is NOT good for crowds unless you know where and when to execute moves like the area-of-effect knockdown one. Even then you still have a small window to pull off more moves on your target.

KayeArray:
Exactly, plasma fist is good when used properly but disintegrate is good when used by anyone. Executing the right combos takes time, is risky and often requires the help of a stun (like mutate) while spells like fireball, lightning bolt and disintegrate can completely fuck people, stunned or not. Plasma fist is more of a sidegrade to disintegrate except only in good hands (or combo'd with mutate which is 3 spell points total) meaning that it makes sense to be cheaper.

trifornt-tinfornt:
Not to mention if you slip up on any of the combos you can't execute any of them until the "COMBO" bar goes away or you hit another person.

Carbonhell:
that isn't true

Crystalwarrior:
If you screw up a combo it will still try to find the combo string in the COMBO bar, so just start over if you do.

trifornt-tinfornt:
Really? It hasn't ever worked like that for me.

Carbonhell:
For me the current price is ok

Nokington:
Nah one should be fine

Gandalf2k15:
I like the current price

calibraptor:
Current price is fine. It's underused as is

Carbonhell:
denied

PR #2176, opened 2016-04-17
Lowers the priority on templates to -2, lower than Air

TDcoolguy300:
As above.


TDcoolguy300:
Forgot to mention, this should fix pipe net runtiming on templates

PR #2178, opened 2016-04-17
Adds a Reassign Area proc to APCs, fixes template powernets even moreso

TDcoolguy300:
As above.

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -139,6 +139,11 @@
 		spawn(5)
 			src.update()
 
+/obj/machinery/power/apc/proc/reassign_area(turf/loc)
+	if(auto_name)
+		name = "[get_area(src)] APC"
+	area = src.loc.loc:master
+
 /obj/machinery/power/apc/Destroy()
 	apcs_list -= src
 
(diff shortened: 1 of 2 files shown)


Gandalf2k15:
please merge

code/modules/power/apc.dm
```diff
@@ -139,6 +139,11 @@
 		spawn(5)
 			src.update()
 
+/obj/machinery/power/apc/proc/reassign_area(turf/loc)
+	if(auto_name)
+		name = "[get_area(src)] APC"
+	area = src.loc.loc:master
```
Carbonhell, line 145: Why colons,can't you do var/area/honk = get_area(src), area = honk.master

Carbonhell, line 145: Also src reee

Carbonhell, line 145: Nvm about src,this is the only case where using src is needed

code/modules/power/apc.dm
```diff
@@ -139,6 +139,11 @@
 		spawn(5)
 			src.update()
 
+/obj/machinery/power/apc/proc/reassign_area(turf/loc)
```
Carbonhell, line 142: Hol up what is turf/loc for,You're not using it

TDcoolguy300, line 142: Incase you pass a turf in

Carbonhell:
Dangerous but acceptable bugfix,merging

PR #2172, opened 2016-04-16
Fixed belt slot not being strippable with no jumpsuit

Carbonhell:
Fixes #1620

PR #2181, opened 2016-04-17
Fixed being able to equip items to belt on your naked body

JohnGinnane:
Is now consistent with ID slot, pockets, etc.

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -478,12 +478,6 @@
 			if( !(I.slot_flags & SLOT_FEET) )
 				return 0
 			return 1
-		if(slot_belt)
-			if(H.belt)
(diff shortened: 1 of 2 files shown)


Crystalwarrior:
This is intended though...
It was confirmed as a feature all the way back to pre-rebase, where even after rebase it was ported back.
Fix stripping belt items not working instead.

JohnGinnane:
Hm, but how can you put an item on your naked belt slot? For example a crowbar

JohnGinnane:
I can change the show_inv() to show the belt if it's not obscured (similar to uniform for example), instead of being a child of the uniform slot

Carbonhell:
https://github.com/HippieStationCode/HippieStation13/pull/2172
I already did it

JohnGinnane:
Ah okay, whoops

PR #2212, opened 2016-04-28
Unhides the face behind the cluwne mask

TDcoolguy300:
They deserve to be seen for their horror


trifornt-tinfornt:
Wait, what does this actually do?

TDcoolguy300:
It doesnt hide the cluwne's face, AKA it prevents the UNKNOWN bullshit with their mask

trifornt-tinfornt:
Ah.
I already suggested this, but Cluwnes in general should just be treated like how they are on Goon, in other words, makes their name "cluwne" which can't be disguised at all.
I think that'd do better here too.

TDcoolguy300:
That requires a bunch of snowflake coding mate, i aint doing that

trifornt-tinfornt:
Regardless, is gud PR.

Lawligagger:
gud

myontasm:
Yeah hiding the face with a clown mask doesn't make sense when the parent doesn't hide it.

TheOneAndOnlyCreeperJoe:
the parent does hide it.

Seems nice to me.

TDcoolguy300:
As creeperjoe said, the normal and sexy clown masks do hide the face, this one shouldnt however.

Nokington:
:+1:

Crystalwarrior:
anyone codechecked this????????

:^)

PR #2218, opened 2016-04-28
Adds more robes to Wizard's Den

TheOneAndOnlyCreeperJoe:
adds a fuckton of robes because the fashion police demands it. Most are palette swaps though.

changed the starting room of the Wizard's ship so that all the wizard clothes are actually in that room.

nicerack


Crystalwarrior:
Oh nice, Fury's Jester Suit is added

doobiedoo23:
Sure

myontasm:
Yasss

PR #2223, opened 2016-04-28
Nerfs Curse of the Cluwne range, adds an icon for the spell

TheOneAndOnlyCreeperJoe:
Coming to think of it a range of 7 is absolute horseshit.

Lowers it to 3, so people will still hesitate coming close to you, but at least you can't cluwne the hos that's lasering you at the edge of your screen..

Also comes with an icon (it was previously using the shitty barnyard icon)


trifornt-tinfornt:
I only saw the spell at close range, but good GOD it had a range of 7?
I'd just say make it function like Flesh 2 Stone and Ei Nath, where you have to click them.

TheOneAndOnlyCreeperJoe:
There would be no incentive to pick it over Disintegrate if it was melee only.
Rule of gimmick isn't a valid argument when it comes to balancing.

trifornt-tinfornt:
True.

myontasm:
Good stuff. +1

doobiedoo23:
Yee

PR #2224, opened 2016-04-28
Allows all species to become hulk

blobbernaut:
On live, only humans can become hulk due to a line on the code that was added, I think because we didn't have sprites for them. That's a bad reason to not allow other species to become hulk, if anything I can bump up the layer the hulk sprite shows up on so that you can see it easier until we get real hulk sprites for the various races.


McHatters:
Oh no wonder I couldn't hulk myself as a black guy :^)

doobiedoo23:
Sure

Limed00d:
Yeah sure, let's end racism today #lizardlivesmatter

PR #2284, opened 2016-05-07
Goodbye detain mode + various other small tweaks

Crystalwarrior:
Removes detain mode due to it existing solely for that stunlock (which is actually useless 99% of the time and only exists to be a cunt to people)
Reduces stun duration for harmbatoning
Updated cattleprod force to be 7 (from 3) so it's on par with crowbar/wrench
Adds "alternative detective fedora" which replaces the purely cosmetic one in det's locker (which was confusing, stupid, dumb and stupid)

### DIFF
--- a/code/game/machinery/bots/ed209bot.dm
+++ b/code/game/machinery/bots/ed209bot.dm
@@ -126,7 +126,6 @@ Auto Patrol[]"},
 "<A href='?src=\ref[src];operation=idcheck'>[idcheck ? "Yes" : "No"]</A>",
 "<A href='?src=\ref[src];operation=weaponscheck'>[weaponscheck ? "Yes" : "No"]</A>",
 "<A href='?src=\ref[src];operation=ignorerec'>[check_records ? "Yes" : "No"]</A>",
-"<A href='?src=\ref[src];operation=switchmode'>[arrest_type ? "Detain" : "Arrest"]</A>",
 "<A href='?src=\ref[src];operation=declarearrests'>[declare_arrests ? "Yes" : "No"]</A>",
 "<A href='?src=\ref[src];operation=patrol'>[auto_patrol ? "On" : "Off"]</A>" )
 	var/datum/browser/popup = new(user, "autoed209", "Automatic Security Unit v2.6")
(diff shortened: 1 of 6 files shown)


Nebyab:
Is the alternative detective's fedora armoured? If so I'm gonna need stats on that.

Otherwise its all good.

Carbonhell:
Detain mode is useful for AI, i use it a lot when playing as it. I personally don't look forward removing it.

Carbonhell:
And on a sidenote you only removed the way to activate detain mode,you didn't actually remove it,unsure if intended or.

Crystalwarrior:
@Nebyab Same stats as normal det fedora.

Crystalwarrior:
@Carbonhell Intended if we have detain mode-only beeposkis or whatever.

Regarding it's use during AI, it's still kinda bullshit on the receiving end since it's an unavoidable stunlock you can never hope to escape from. Worst part is if they forget about you you'll just be stunlocked for the rest of the round.

dreamcrab:
should harmbatoning really reduce the stun time?

Crystalwarrior:
Sure, I mean that implies you're not supposed to use harmbaton as a stunlock
@Nebyab came up with the idea, in his (not exact) words: "Stuns with damage should compensate with less stun time, like the detective revolver"

carlarc:
Carbonhell, yes.
Detain mode is worse than death. You're stuck there until roundend unless it's malf.

Sarcalogo:
I was once stunlocked for 15 in real life minutes, endlessly calling out for help, as the Captain in a Malfunctioning AI round. The AI core was ravaged by the singularity, but Pingsky survived and stunned me forever. I died because I ran out of air. I was FURIOUS at Overlop Funk for leaving me to die in this situation, who about 13 minutes in wizzed past me in an eva suit while only 4 tiles west of me, with my antique laser gun on the ground, ready to kill Pingsky. The AI won in the end.

dreamcrab:
im not entirely opposed to the harmbaton change, i just wanted to see why it was happening. i can accept that logic.

but detain really does suck. you need to be saved by someone or youre stuck forever. should we leave in detain but then punish AI players or whoever set it to detain for that? or just remove it? im not against removing it but maybe making beepsky a bit more tanky

Crystalwarrior:
Removing it is a much better option than forcing arbitrary administration policies on admins, methinks.

Lawligagger:
seems good to me

ghost:
Yeah it's good

Crystalwarrior:
Merge it then :^)

PR #2261, opened 2016-05-02
Shadowling Glare nerf

ghost:
Glare and Lesser Glare have no effect on targets that are blind. (It still goes on cooldown as you attempt to blind them.)
Lesser Glare is now countered by anything that gives eye protection (Welding goggles/masks, sunglasses and sechuds.)

PR #2304, opened 2016-05-10
Shadowling nerfs boogaloo fuck

blobbernaut:
Ascendants now punch things at lightspeed, and do half damage per punch. This is a DPS increase of 250%.

Shadowlings heal in the dark at 3.5 units of each damage type per tick instead of 5.

Super thralls heal for 3 now instead of 2.

Shadow walk is now 800 tick cooldown instead of 300, does not remove stuns, and does not make the user fully invisible, and lasts for only 3 seconds instead of 4.

Icy veins now only has a range of 3, from 5, and cannot stun.

Re-hatch now also removes stuns, and has a cooldown of 120 seconds. The second time you catch a changeling after they jump, they will have to fight you. 

Sonic screech now has a range of 5, to fit with drain life, which now has a range of 5, from 3.

Both glares now have a range of 3, from fullscreen.

Shadow walk makes the thrall fully invisible, to make it actually worth using. 

Adjusted the ranks at which shadowlings acquire new powers to fit in flash freeze at 8 thralls acquired, buffs it to make it inject more frost oil.

Shadowlings now start with knock to compensate for the longer cooldown on jaunt. 

Oh this also fixed a bug where the chitin helmet was acting as a firesuit helmet.


trifornt-tinfornt:
80 second recharge is overkill. Also "does not make the user fully invisible" kinda destroys the element of stealth and moving from maint to maint to not be detected. And no, Knock is not a worthy compensation (although it does make a lot of slings without maint access' lives easier).

Also Icy Veins was ever able to stun? When? Thanks for removing that if it actually did exist.

Thank you for buffing superthralls though. There's literally no reason to not use it.

blobbernaut:
Then what do you suggest? Some sort of mechanic where they can't jump if they're taking damage? I think 80 is fine. Shadowlings shouldn't be able to avoid any and all fights because they can jump around on no cooldown. 

Oh this also fixed a bug where the chitin helmet was acting as a firesuit helmet.

trifornt-tinfornt:
Thing is you're making it so that jaunting will immediately reveal the fact that Slings exist, and then nerfing the ability more on top of that.
It's really dumb and unnecessary. Nobody even likes Slings (myself included) so I'd just say remove the gamemode - don't waste your time balancing it. And if you are gonna try and balance it don't make it kind of ludicrous.

blobbernaut:
it'd be easy as fuck to remove the gamemode, it's just a config option. Tell the headmins when you see them in game, tell them on the forums, but this is for attempts are balance.

trifornt-tinfornt:
All right.
Personally the best balance would be
"Removes Shadowlings."

dreamcrab:
yeah, not exactly feeling this. there might be ideas in here we could use if slings are fucked with, though

blobbernaut:
We aren't NOT changing shadowlings. I'm going to make PR's until one of them is accepted and shadowlings get nerfed hard.How can you possibly think they're in an okay spot right now?

blobbernaut:
ideally i'd just want shadowlings removed.

SugarBooger:
Wtf don't remove sligs they're fun

doobiedoo23:
ITPR: blobber got killed by shadowlings and started grudgecoding, is pissed off to the point that he'll continue making PRs if this one doesn't get accepted

McBawbaggings:
Ascendant change is just punting numbers around for no reasons, but fist of the north star ascendants seem cool.
Heal change seems okay. Realistically, it's only another 8 ticks and a half to fully heal from crit (0 HP)
Super thrall change is okay
Shadow walk nerf is good BAR removing full invis (You changed thralls to be fully invis but slings don't?) (Fix the comment please too, it says 4 secs)
Icy veins nerf is legit.
Re-hatching is actually useful now.
Is the glare nerf still required after I added protection against lesser glare?: https://github.com/HippieStationCode/HippieStation13/pull/2261
Repeating myself: Why do thralls get full invis and slings don't? Even then, imagine trying to catch a invis fucker. I get they don't get free movement like slings but still.
You're saying you've buffed icy veins to have more frost oil when both the codebase and your PR have it inject 15 frost oil? Even then, isn't 15 enough to crit you without coffee anyway? Coffee would remove it even if it was like 30 units or something dumb.
Knock change is great, fuck being medbay doctor and sling.

carlarc:
I really want the number of needed thralls increased, you barely get a chance to test out unlockeable powers such as superthrall because oops, 1 thrall away from ascending

PR #2313, opened 2016-05-13
Armor update!

blobbernaut:
Armor now stacks multiplicatively!
For example, an 80 armor riot suit and a 10 armor sec jumpsuit combine for 82 armor now, not 90!

Cloaks! Cloaks are low storage, and can be worn on the back(!) or the exo suit slot. On the back, they'll have 15 slots, 5/7ths that of a full backpack, but have a bit of armor to go along with it. Thanks to the multiplicative armor part they won't stack to infinite armor. In the exosuit slot, they'll have a bit less armor than regular vests, but grant a bit of storage space!

Labcoats, on a completely unrelated note, now have some protection against disease. The traitor labcoat gives the same disease protection as the bio suit! Go get em, you rascally virologist you! 

Accessories now have armor!
Who knows what we can do with this. Bulletproof waistcoats? SURE! protective talismans for cult? Why not? It's all possible now. I have updated it so that a medal will give you 10% bullet resistance! However, with the captain's carapace, that's only 6% as it starts with 40! Guess what that does? That's right, you can now survive 3 revolver shots, assuming they aren't aiming for your face, and live with a whole 2.8 HEALTH! Assuming none of them hit your legs or arms, that is!

It might be a good idea to buff jumpsuit based armor. Adding some laser and a lot more melee protection on the sec jumpsuit wouldn't be that bad.


blobbernaut:
only thing to talk about is armor numbers, now.

McBawbaggings:
Isn't this moreso a nerf to a lot of small bonuses since it will be repeated diminishing returns rather than 1 small diminishing return?

What's currently broken about the current armor system that makes it utter garbage? Aren't the weapons currently balanced around the additive system?

Do biosuits now not give 100% bio resistance towards diseases?

Accessories having armor is nifty though.

blobbernaut:
What do you mean by repeated diminishing returns? Say you equip an item that has 50 armor. This will reduce 50% of the incoming damage that is not blocked by the rest of your armor. 

and additive it isn't garbage, it's just that additive values make it hard to actually make jumpsuits or back cloaks have reasonable armor values on their own because they'd stack so quick to 100%. An example of this is gang jumpsuits and the bulletproof vest giving you 100% bullet protection on your chest. 

Right now, it's a nerf to both jumpsuits that give armor (which is currently the ONLY thing that we have that stacks armor, and thus the only thing affected by this) but without this change, we'd have: Jumpsuits, accessories, vests, and back slots. And you could very quickly get stupid high protection, even if the items themselves aren't strong on their own, and that's what I want to avoid. 

To compensate, security jumpsuits could have their armor increased to 20 or so, which would make it the same on the chest, and higher on the limbs. Gang jumpsuits are probably fine, I don't think people realized quite how tanky those made you. 

I just think it's a neccessary change if we add in backslot armor that is good standalone, otherwise, it'll stack too quickly. HOS coat + flat armor cloak + flat armor jumpsuit would just be absurd unless the numbers on the jumpsuit and cloak were so low as to be almost negligible without the coat.

blobbernaut:
sorry if that's jumbled and incoherent, I'm very tired right now

McBawbaggings:
Like the value you gave right, 2 pieces of armor. 1 is big and the other is small.
If they were evenly distributed resists like 45% It'd be 100 * 0.45^2 which is 20.25 so you'd get about 80% resist as opposed to 82%.
So 1 big armor + 1 small armor is still better than 3 medium armors if their additive percentages are the same:

Armor 1: 60% coat + 5% jumpsuit = 62%
Armor 2: 20% coat + 20% jumpsuit + 5% medal = 38.2%
Old armor system for both is 65%

McBawbaggings:
Fixed muh maths.

blobbernaut:
the medal just increases the jumpsuit's defense because it's not a slot in of itself, the only area where three pieces of clothing can guard at once is if you're wearing a cloak, an exosuit slot armor, and an armored jumpsuit. 

Let me make it clear that armor stacking is not actually a thing right now. The only thing that uses it is gang jumpsuits and the 10 melee armor on security jumpsuits. However, that could be changed, and you're free to suggest changes to the various armor values of jumpsuits and the cloaks.

McBawbaggings:
I'm just saying that there will probably be have to calculations of the general set up of things since lots of small armors on a multiplicative system is objectively worse than 1 or 2 armor slots that are more efficient. 

Anything midtier in defense stats is probably gonna get hit the hardest since you'll probably be packing multiples of that sort of thing.

McBawbaggings:
Considering I'm dumb and jumpsuits generally speaking lack stats, you're probably not going to have that scenario I described. Do we still get full bio resist though?

blobbernaut:
Yes, and that's intended. A lot of small armor is supposed to be weaker than one piece of armor that has armor equal to the sum of their stats. I think that's a bit more intuitive. 

Also i'll have to check the code for bio armor but I don't think anything uses that to check for damage so I don't think so

blobbernaut:
Yes, and that's intended. A lot of small armor is supposed to be weaker than one piece of armor that has armor equal to the sum of their stats. I think that's a bit more intuitive. 

Also i'll have to check the code for bio armor but I don't think anything uses that to check for damage so I don't think so

blobbernaut:
I haven't found a single thing in the code that actually USES bio armor. I think the bio suit's protection comes from the fact that it covers your whole body and has permeability = 0.01

carlarc:
yes

Nebyab:
someone explain to me why we needed an armor overhaul at all

blobbernaut:
why do we need any features? 

but this is not even an overhaul. The only things this affects currently is the sec jumpsuit, and gang jumpsuits. Sec jumpsuit's 10 melee armor is barely noticeable anyway.

So you could hardly call this an overhaul. This will just allow for the existence of armored jumpsuits or backpack slot items in the future and allow them to be strong on their own but not cause invincibility when combined with a strong armor item.

PR #2338, opened 2016-05-15
Readds the gateway once again.

Gandalf2k15:
This re adds the gateway once again.
http://hippiestation.com/showthread.php?tid=3993


TDcoolguy300:
This should be fine, merging as it was already approved

PR #2341, opened 2016-05-15
Lings cannot use gateways

blobbernaut:
This was in another PR with mixed to unfavorable opinions. I could remove the ability for this to be used as a ling test by: 

*Lings lose their changeling powers when in the gateway, period, or at least, remove their ability to do anything except regenerate 

However, in response to that, I just think that a ling's abilities, such as adrenaline surge, combined with the loot they can get is the problem. It's not a matter of them clearing the away mission or getting free guns, the problem is when they come back to the station, they are a changeling, an already formidable foe, with access to things such as demon hearts, and deathsquad armor, and assault rifles.

*Have the teleportation be done by clicking on the gate, turning the portal off or on is done by a panel or button.

If you are a headmin, and vote no on this PR, please also tell me if you want one of those other options added.

___

### DIFF
--- a/code/modules/awaymissions/gateway.dm
+++ b/code/modules/awaymissions/gateway.dm
@@ -141,6 +141,11 @@ var/obj/machinery/gateway/centerstation/the_gateway = null
 		return
 	if(!active)
 		return
+	if(ismob(AM))
+		var/mob/M = AM
+		if(M.mind && M.mind.changeling)
+			M << "The ancient builders of this gate seem to have barred our kind from using their technology..."
+			return
 	if(!awaygate || qdeleted(awaygate))
 		return
 
@@ -252,6 +257,11 @@ var/obj/machinery/gateway/centerstation/the_gateway = null
 		return
 	if(!active)
 		return
+	if(ismob(AM))
+		var/mob/M = AM
+		if(M.mind && M.mind.changeling)
+			M << "The ancient builders of this gate seem to have barred our kind from using their technology..."
+			return
 	if(!stationgate || qdeleted(stationgate))
 		return
 	if(istype(AM, /mob/living/carbon))


blobbernaut:
Of note is that pushing a ling against the portal does not give any notice or visual effect to anybody other than the ling. It will just look as though they cannot be pushed inside. If you ask my opinion, due to the lack of actual feedback, it could also just be a rule to not use this as evidence of ling-ness.

Carbonhell:
Or perhaps transform the gateway in some sort of VR?So what you are before entering doesn't matter,ie traitor items won't make your life easier aswell

Crystalwarrior:
One, lings can no longer eat corpses that were never "human" in the first place.
Two, I find that this isn't the issue exclusive to changelings as traitors and any other antagonists can exploit the gateway for that SWEET OVERPOWERED LOOT, so this is more like a patchwork for changelings rather than an actual solution to the problem.

I dunno why we can't just put the gateway itself on another space Z-level so you have to at least do some space exploring before you can either man the gateway yourself as a lone wolf or set up a teleporter to bring the crew over for the fun.

blobbernaut:
So tl;dr; My main reason for making this PR is that changelings with access to the guns they get on the missions are too strong AFTER COMING BACK, so that's why I want to bar them. 

Let's go in a list by list of the antags that can use the guns and stuff on away missions:

Traitor: Benefits much less than lings. They can get holoparasites for healing, or adrenals for combat when they return, but their items give them limited combat capability while there. They have no natural healing so they are as likely to die as other crew unless they buy a support holo.
The thing about adrenals is that it's temporary. 

Wizard: Mostly will be crew members on here. No problems here, crew usually has guns and spells.

Shadowling: Shadowlings won't be exploring much, since shadowlings CANNOT USE GUNS or wear armor. 

Gang: If a gangster wants to raid the away missions, that's fine. They can already get uzis and shit, and they can't make a hideout here.

Revs: Both head revs and heads of staff count as dead if they leave to the gateway so they won't be coming. 

Nuke ops: Could see use for the crew getting guns. What the hell is a nukie gonna be doing raiding the away missions? 

Abductor: No reason to get here, abductors rarely if ever murderbone. 

Cultist: Cult rounds usually last a while, and I believe teleport runes work off station. They have quite a few useful abilities such as raise dead. (Note: Having an event where the whole station is cult and they explore a super hard away mission is a fucking great idea holy shit let's do that sometime.)

blobbernaut:
Now lings ( I'll update this post, I just wanted it to be directly after the previous one but in a seperate post because it'll be long)

Crystalwarrior:
I still think that even Traitors could have some use out of the gateway. Not to mention you don't neccesarily have to be a revhead to get that sweet OP loot.

Gateway in its current form breaks balance in a lot of places, and I think that not only do we need to make gateway missions less of an unbalanced mess, but I also think that GETTING to the gateways is a bit TOO easy. There's no effort involved, just straight up waiting.

calibraptor:
I'd honestly support the delay being upped to 15 minutes, so the ling either has to be very skilled or end up coming back to the station when the round is likely on the verge of ending.

carlarc:
Maybe rework the gateway so it's "bluespace headsets" or some yadda yadda, where you control a (gameplay-wise human) robotic body that can't leave the away mission

yobob591:
while thats not a bad idea @carlarc I think at least half the reason the gateway exists is for the loot, turning it into VR or remote control would have you end up with 1 or 2 people in it each round like Goon's VR.

carlarc:
b-but muh earthbound references

Pyko1:
What if when entering the gateway as ling you get this message.
"The gateway seems to be unable to properly hadle our genetic structure, our powers are diminished."
Which could mean no powers at all, or only some which we feel suitable.

And upon returning to station you get.
"Our full genetic powers has been restored from gateways internal buffer."
Meaning everything is restored to normal.

Ultimately turning the ling in to a human in gateway or to something almost as weak.
I think regen should be disable at least in gateway, as you could die die die die die, get loot kill everything.

carlarc:
even as human you'll get gratuitous amounts of loot that can be used in the station.

i don't like the click to teleport, you can't pull medibots into the gateway that way.

Pyko1:
Also rule against ling testing with gateway is kinda iffy.

if someone tries to just toss someone in gateway for lols, or exile them. Do we ban them for accidentally ling testing? And the people actually ling testing, but claiming they were doing the before.

Better just have it work as it did before, and deal with the ling powers that are too good in gateway.
And yeah the click thinggy makes exile implants useless, unless you can convince people to exile themself.

Crystalwarrior:
Still don't understand why we should specifically butcher changelings on gateways instead of having gateways that can't be easily cheesed with ling powers. :v

carlarc:
because that'd be boring for hoomans

Pyko1:
Because changelings can always die die die die die die die, oh my phat loot kill errywing.

Unless you have every mob gib you when killed.

TDcoolguy300:
Code in theory would work

McBawbaggings:
Outright preventing changelings going through gateways is a bad idea.
To be honest, I'd rather gateway stuff was less loot and more diversion based which would make lings going into gateways not as dangerous.

doobiedoo23:
This'd become a way of testing if someone is a ling.

dreamcrab:
gateways are dead for now anyways, kill this.

Kokojos:
Better luck next time

PR #2347, opened 2016-05-16
Supermatter fixes

calibraptor:
- Added actual areas for the new engineering rooms.
- Made ridiculously efficient vents less ridiculously efficient, also added volume pumps to the scrubbing pipes
- Supermatter shard can now be ejected from the core, and won't be pushed around by air.
- Didn't use vacuum floors in the SM core because I'm an idiot. Fixed now though.


calibraptor:
Wait what WHY ARE THERE NO FILES

Kokojos:
Seems right. merging since this is a fix.

PR #2350, opened 2016-05-16
PROPER zombie fix (+ Fixes an issue with diseases)

Crystalwarrior:
Fixed an issue with diseases not being properly removed from the mob virus list if disease is irresistible
Allowed zombie gamemode to be a roundstart gamemode if forced by admins

### DIFF
--- a/code/datums/diseases/_disease.dm
+++ b/code/datums/diseases/_disease.dm
@@ -156,7 +156,7 @@ var/list/diseases = typesof(/datum/disease) - /datum/disease
 		if(disease_flags & CAN_RESIST)
 			if(!(type in affected_mob.resistances))
 				affected_mob.resistances += type
-				remove_virus()
+		remove_virus()
 	qdel(src)
 
 
--- a/code/datums/diseases/zombie.dm
+++ b/code/datums/diseases/zombie.dm
@@ -99,8 +99,8 @@ var/list/zombie_cure = list()
 	else
 		if(stage >= max_stages && ishuman(affected_mob))
 			var/mob/living/carbon/human/H = affected_mob
-			H.Zombify()
 			cure()
+			H.Zombify()
 			return
 		if(stage_prob <= initial(stage_prob))
 			stage_prob *= 2
--- a/code/game/gamemodes/zombie/zombie.dm
+++ b/code/game/gamemodes/zombie/zombie.dm
@@ -21,10 +21,6 @@
 
 	var/players_per_carrier = 11 //1 patient zero every 13 players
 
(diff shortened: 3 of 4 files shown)


Gandalf2k15:
Please merge asap.

code/datums/diseases/zombie.dm
```diff
@@ -99,8 +99,8 @@ var/list/zombie_cure = list()
 	else
 		if(stage >= max_stages && ishuman(affected_mob))
 			var/mob/living/carbon/human/H = affected_mob
-			H.Zombify()
 			cure()
```
JohnGinnane, line 102: Does this mean we cured people immediately after cutting them?

Crystalwarrior, line 102: No, this is actually just a failsafe in case affected_mob disappears during Zombify() (which still shouldn't happen but w/e)

code/datums/diseases/_disease.dm
```diff
@@ -156,7 +156,7 @@ var/list/diseases = typesof(/datum/disease) - /datum/disease
 		if(disease_flags & CAN_RESIST)
 			if(!(type in affected_mob.resistances))
```
JohnGinnane, line 157: Where is type defined?

code/datums/diseases/_disease.dm
```diff
@@ -156,7 +156,7 @@ var/list/diseases = typesof(/datum/disease) - /datum/disease
 		if(disease_flags & CAN_RESIST)
```
JohnGinnane, line 156: Won't this mean that you'll remove the virus even if they can resist it?

code/datums/diseases/_disease.dm
Crystalwarrior, line 157: disease itself probably? This is old code

code/datums/diseases/_disease.dm
Crystalwarrior, line 156: That's what is SUPPOSED to happen. CAN_RESIST flag is only suppsoed to add the cured disease to resistances so you are immune to it next time.

JohnGinnane, line 156: Ah yes that makes sense!

code/datums/diseases/_disease.dm
JohnGinnane, line 157: Nope, it seems type is a read only variable which returns the type of object, handy enough

Gandalf2k15:
MERGE

JohnGinnane:
Code looks good, merging

PR #2351, opened 2016-05-16
Disables the gateway

Gandalf2k15:
Disables the gateway until further notice. Its lagging for some unknown reason.

### DIFF
--- a/code/modules/awaymissions/awaymissionconfig.txt
+++ b/code/modules/awaymissions/awaymissionconfig.txt
@@ -9,6 +9,6 @@
 
 #_maps/RandomZLevels/wildwest.dmm
 #_maps/RandomZLevels/challenge.dmm
-_maps/RandomZLevels/moonoutpost19.dmm
-_maps/RandomZLevels/caves.dmm
+#_maps/RandomZLevels/moonoutpost19.dmm
+#_maps/RandomZLevels/caves.dmm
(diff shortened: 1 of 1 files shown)


blobbernaut:
RIP :(

blobbernaut:
are you sure it's not the zombies

Gandalf2k15:
I'm sure
If it was it would be diseases.

blobbernaut:
what was lagging specifically

Crystalwarrior:
mob controller, I still suspect it's zombies.

Gandalf2k15:
Mobs was at 5.

Crystalwarrior:
Hm, actually I dunno. It might be gateway just as it might be zombies.

blobbernaut:
yeah there were no more living mobs on the gateway.

Gandalf2k15:
Disable it just to be sure I guess.

JohnGinnane:
Merging to confirm whether it's still zombies or potentially gateway.

calibraptor:
Gateway will be brought back later right? I still want to make maps for it. :(

JohnGinnane:
I believe that is the intention. Disabled temporarily while we look into
what's causing the lag as it's unbearable and can crash the server

blobbernaut:
blease revert

PR #2358, opened 2016-05-16
Re-enables the gateway (ubermeat urgent)

Gandalf2k15:
Its not the gateway causing the lag anymore. The caves map was causing
major problems, removes this.
Also this is an extremely important PR as the server will not restart correctly without maps in the config. Merge ASAP.


JohnGinnane:
Will merge as soon as Travis finishes.

JohnGinnane:
Travis is happy, so merging

blobbernaut:
thank

Lawligagger:
why you do

Gandalf2k15:
I had originally though it was this, but it was not. Apparently another fix broke the server. Separate from the gateway.

PR #2365, opened 2016-05-17
Restrart bug finding 1 - Zombies

Pyko1:
commented out most of the zombie stuff.

Pretty sure its not this tho.


Crystalwarrior:
Code's OK, though I'm still thinking it has nothing to do with zombies.

Pyko1:
as do, I just rulign it out, mergine as this is a fix and code is checked

Issue #2364, opened 2016-05-17
Server crashes every time it tries to restart, requiring manual restart


calibraptor:
Just thought I'd make this an official issue. Hop on it code monkeys, chop chop.

JohnGinnane:
What's happening when you try to restart?

calibraptor:
@JohnGinnane 

At the end of the round, when the server tries to automatically restart for a new round, at some point inbetween the round ending and the start of the next round, it crashes. GMs/Headmins have been having to manually restart it each time.

JohnGinnane:
Can we view the output of the dream daemon or is there some way of knowing what's happening?

blobbernaut:
I think TD said the garbage collector was fucking up

JohnGinnane:
Okay, will do some local testing tonight to see what I can replicate

Pyko1:
Either zombies or gateway, or the server is shitting itself.

JohnGinnane:
Tested locally and appears to reboot, but plenty of errors that need to be fixed I think.

[server_log.txt](https://github.com/HippieStationCode/HippieStation13/files/269073/server_log.txt)

PR #2369, opened 2016-05-17
Added temporary debugging messages for the server

JohnGinnane:
Just added some simple debugging. Perhaps another coder in another time zone could check this out?


nalarmg:
It would be better to change these to the message admins proc or something else not visible by players.

JohnGinnane:
Good point, hold on and I'll make that change and test it..

JohnGinnane:
Done! I'm heading to bed in the next hour though so if someone does merge and server is updated, would someone mind checking the output of this and logging under #2364

Carbonhell:
not needed anymore i assume?closing

PR #2372, opened 2016-05-18
New Zombies + MUTATIONS! (READY)

Gandalf2k15:
These zombies DO NOT OVER WRITE ANYTHING. They are a completely new mob, they do not use a virus to spread. That means no stress on the disease controller which means no lag. These are almost identical in how they function. I have tested these at least 100 times. They have never lagged. They shouldn't anyway. Code check please too.

The system is infection system is in several stages, takes a random amount of time for some to turn.

People are infected via bites, which has a random chance to happen when you attack someone. Once bitten there is no cure. _Yet._

Also I plan on adding mutations that you can purchase. Said currency will be based off of how many people you infect.
-DONE-
Now you have a system where you can upgrade your zombie!
All zombies start with 5 points.
You gain 3 
The abilities are:
- Evolve Health(cost: 1): Will increase the max zombie health by 10 and restore 10 (cap 300)
- Evolve Attack(cost: 1): Will increase the max and min damage by 2 (cap 60)
- Evolve Speed(cost: 1): Will increase the max(min) speed by 0.1 (cap -2)
- Self Revive(cost: 7): When purchased the zombie will auto regen but will lose all prior stats
- Force Door(cost: 3): Gives the ability to force open doors (starting 20 seconds)
- Airlock Force Time(cost: 1): Gives the ability to force open doors faster (cap 2 seconds)
- Super Form(cost: 10): Makes you invisible(2000HP), buffs stats for a certain amount of time, then it wears off.
- Upgrade Super Form(cost: 2): Increases the amount of time you are in super form. (cap 30 seconds)

Screenshot of mutations tab:
</img>
</img>


Gandalf2k15:
KEK forgot files

Zadeon:
:+1:

trifornt-tinfornt:
Is this the Plague Inc. idea I suggested a while back?

Gandalf2k15:
aye

Gandalf2k15:
Still pushing mutations, gonna take a bit as its in testing.

Gandalf2k15:
added mutation info

carlarc:
add 63.75 or whatever chance to be immune to the virus due to WASHING HAND POWER

Gandalf2k15:
wooooops forgot includes

Gandalf2k15:
Mutations pushed, ready to merge.

ghost:
No mergerino until headmin approval and after the revert.

Gandalf2k15:
DO THE REVERT NOWW

ghost:
Infect verb should be under secrets (What permissions do you need to even have to use infect?) (It should also probably be randomly selected among living players)
Disease itself works way too fast and the disease progression is a bit dry (Could use more span classes and blood vomiting)
Change "You dont feel right! Somethings wrong!" to "You don't feel right! Something is wrong!"

Biosuits and hardsuits should give you immunity to infection. (I like the idea of it being incurable, but there needs to pre-emptive measures to block it. Maybe also make it on the first bite, the victim has a tiny chance to become immune.)

Speed upgrades should be rounded and probably represented in percentages (Currently as is, it looks pretty bad when it starts representing stuff in scientific notation.)
Infecting someone should probably be the nice blue text rather than a warning, and keep the bites as normal red.
Change "That bite felt sore as hell! Its getting worse..." to "That bite felt sore as hell! It's getting worse...." (Or do a spawn and make the bite sound like it's beginning to fester and decay.)

Superform seems way too powerful, (2000HP and 100 dmg attacks?) even if it is expensive. (The lack of sprite isn't good either)
(Code wise, it looks to me that you can upgrade in superform and then lose the changes? Even worse, be locked out of speed upgrades if you were to chug buy them all during it.)

Airlocks should probably be pried open, powered airlocks should be slow and unpowered ones should be quick.

Gandalf2k15:
aight, well I'll make these changes. But I really want it for an event I have planned too.

PR #2377, opened 2016-05-19
New Zombies + MUTATIONS! (READY) #2372

Gandalf2k15:
**THIS IS ONLY FOR ADMINBUSSING AT THE MOMENT**

These zombies DO NOT OVER WRITE ANYTHING. They are a completely new mob, they do not use a virus to spread. That means no stress on the disease controller which means no lag. These are almost identical in how they function. I have tested these at least 100 times. They have never lagged. They shouldn't anyway. Code check please too.

The system is infection system is in several stages, takes a random amount of time for some to turn.

People are infected via bites, which has a random chance to happen when you attack someone. Once bitten there is no cure. _Yet._

Also I plan on adding mutations that you can purchase. Said currency will be based off of how many people you infect.
-DONE-
Now you have a system where you can upgrade your zombie!
All zombies start with 5 points.
You gain 3 
The abilities are:
- Evolve Health(cost: 1): Will increase the max zombie health by 10 and restore 10 (cap 300)
- Evolve Attack(cost: 1): Will increase the max and min damage by 2 (cap 60)
- Evolve Speed(cost: 1): Will increase the max(min) speed by 0.1 (cap -2)
- Self Revive(cost: 7): When purchased the zombie will auto regen but will lose all prior stats
- Force Door(cost: 3): Gives the ability to force open doors (starting 20 seconds)
- Airlock Force Time(cost: 1): Gives the ability to force open doors faster (cap 2 seconds)
- Super Form(cost: 10): Makes you invisible(2000HP), buffs stats for a certain amount of time, then it wears off.
- Upgrade Super Form(cost: 2): Increases the amount of time you are in super form. (cap 30 seconds)

Screenshot of mutations tab:
</img>
</img>


ghost:
~~Infect verb should be under secrets (What permissions do you need to even have to use infect?) (It should also probably be randomly selected among living players.)~~
~~Disease itself works way too fast and the disease progression is a bit dry (Could use more span classes and blood vomiting)~~
~~Change "You dont feel right! Somethings wrong!" to "You don't feel right! Something is wrong!"~~ "You dont feel right" still needs changed.

Biosuits and hardsuits should give you immunity to infection. (I like the idea of it being incurable, but there needs to pre-emptive measures to block it. Maybe also make it on the first bite, the victim has a tiny chance to become immune.)

Speed upgrades should be rounded and probably represented in percentages (Currently as is, it looks pretty bad when it starts representing stuff in scientific notation.)
~~Infecting someone should probably be the nice blue text rather than a warning, and keep the bites as normal red.~~ The victim should get the user danger text, the zombie mob should get blue text.
~~Change "That bite felt sore as hell! Its getting worse..." to "That bite felt sore as hell! It's getting worse...." (Or do a spawn and make the bite sound like it's beginning to fester and decay.)~~

~~Superform seems way too powerful, (2000HP and 100 dmg attacks?) even if it is expensive. (The lack of sprite isn't good either)~~ Still pretty stronk, not sure how I feel about it.
~~(Code wise, it looks to me that you can upgrade in superform and then lose the changes? Even worse, be locked out of speed upgrades if you were to chug buy them all during it.)~~

~~Airlocks should probably be pried open, powered airlocks should be slow and unpowered ones should be quick.~~

Reposting my criticism here.

Gandalf2k15:
Already amended these.

ghost:
Oh righto

ghost:
Updooted my criticism. Lemme know if I missed something.

Gandalf2k15:
(Code wise, it looks to me that you can upgrade in superform and then lose the changes? Even worse, be locked out of speed upgrades if you were to chug buy them all during it.) fixed this

Gandalf2k15:
Infect verb should be under secrets (What permissions do you need to even have to use infect?) (It should also probably be randomly selected among living players.) Perms are same as BSA, functions like BSA too.

Gandalf2k15:
Biosuits and hardsuits should give you immunity to infection. (I like the idea of it being incurable, but there needs to pre-emptive measures to block it. Maybe also make it on the first bite, the victim has a tiny chance to become immune.) Not entirely sure how to do this.

Gandalf2k15:
Still pretty stronk, not sure how I feel about it. Considering it only last 10 seconds and costs 10 is pretty balanced. Its supposed to make you unbeatable for that amount of time.

Gandalf2k15:
Speed upgrades should be rounded and probably represented in percentages (Currently as is, it looks pretty bad when it starts representing stuff in scientific notation.) Forgot about this

yobob591:
I think if the zombie can bite through a biosuit or something it doesnt make sense to make you immune. Hardsuits, yes I agree, you should need some sort of upgrade to bite through metal

carlarc:
honestly anything with melee damage reduction over OR 15 should make you immune to normal biting from that part. if the zombies really want to infect someone they can strip then

yobob591:
I dont know about zombies being able to strip people. Depends on what style of zombie these are, but most traditional zombies are either too dumb or dont have the dexterity needed to take off peoples clothes.

yobob591:
Id go for the evolve strength stuff improving bite power to get through stronger materials

carlarc:
xenos can strip

Gandalf2k15:
Yeah, hardsuits will be an evolveable power

Gandalf2k15:
But I'm not getting into all that armor level buyllshit

yobob591:
xenos are suprisingly dexterous and according to canon the queen is around as smart as a normal human, but you have a point

Gandalf2k15:
This is better than the current zombies by a mile.

carlarc:
well we can't enforce zombie rp now can we

yobob591:
no, but mechanic wise if the zombie is smart enough to remove clothes is the debate here. like how people who are brain damaged just get the message "x stares drools at/cluelessly at the x" or "you momentarily forget how to use x"

realism or no zombies using computers would be weird

yobob591:
but are we just zombies on our computers????

Gandalf2k15:
Its fine as is. I'm not gonna change the hardsuit shit. I don't even know how to do that. I'll fix the current bugs. Remember this is an adminbus command.

yobob591:
Ah, i think it could work as a really rare event/midround antag like memeductors or ayylmaos though

Gandalf2k15:
That may come aye.

calibraptor:
If this'll be a rare midround thing then +1

yobob591:
Makes sense to make it admin spawn only for now though since its just testing

Gandalf2k15:
I'll see about making this a rare midround like ayys

Gandalf2k15:
Aight this is an adminbus thing for now.

yobob591:
meeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeerge

Zadeon:
Better fix the text saying <color='blue'> in chat and stuff.

Gandalf2k15:
aye pushin fix

Gandalf2k15:
I managed to fuck this up somehow. I'll fix it now.

Gandalf2k15:
this fucked up
repring

Gandalf2k15:
Okay abandon ship, I can't fix the fuckups.

PR #2378, opened 2016-05-19
Potential fix for server crashing on reboot

JohnGinnane:
Replaced sleep(delay) with spawn(delay) in an attempt to fix the server crashing at that point, as indicated by debugging showing us that "test 1" is achieved, but not "test 2"

### DIFF
--- a/code/world.dm
+++ b/code/world.dm
@@ -157,44 +157,57 @@ var/global/list/map_transition_config = MAP_TRANSITION_CONFIG
 
 	message_admins("test 1")
 
-	sleep(delay)
+	spawn(delay)
+		message_admins("test 2")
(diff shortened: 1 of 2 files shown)


JohnGinnane:
Did local testing and interestingly the message order towards the end becomes "6, 9, 7, 8" would tells me the Reboot() proc returns and then the spawn(0) is executed afterwards.

I've also added some more "test" messages in case we get lucky and we're able to go further into Reboot().

If someone could review, merge, and update that would be appreciated, but **please be sure the change from sleep() to spawn() wouldn't upset anything else!**

calibraptor:
+1

We need to make some headway on this

PR #2389, opened 2016-05-21
Adds Timelost Spellcasters

TheOneAndOnlyCreeperJoe:
A boss mob that has a combined health of 1000 and is split up into several phases. Will blow up a lot of things in the first, charge at you with his sword in the second and use all of his previous abilities in the hird. Is impossible to defeat on your own even with the entire armor, HoS locker in your backpack.

Primarily used for adminbus and (maybe future) gateway missions.

http://i.imgur.com/FUUjZ6s.png


Kierany9:
yes

but ill have to say no because travis hates you

TheOneAndOnlyCreeperJoe:
Need help on why the fuck it won't find the type path

PR #2408, opened 2016-05-22
Stops from catbanned folks getting that sweet golem rune.

ghost:
Does what is says on the tin.

### DIFF
--- a/code/modules/research/xenobiology/xenobiology.dm
+++ b/code/modules/research/xenobiology/xenobiology.dm
@@ -391,8 +391,9 @@
 /obj/effect/golemrune/process()
 	var/mob/dead/observer/ghost
 	for(var/mob/dead/observer/O in src.loc)
-		if(!O.client && !(jobban_isbanned(O, "catban") || jobban_isbanned(O, "cluwneban")))	continue
+		if(!O.client)	continue
 		if(O.mind && O.mind.current && O.mind.current.stat != DEAD)	continue
+		if(!(jobban_isbanned(O, "catban") || jobban_isbanned(O, "cluwneban")))	continue
 		ghost = O
 		break
 	if(ghost)
@@ -403,8 +404,9 @@
 /obj/effect/golemrune/attack_hand(mob/living/user)
 	var/mob/dead/observer/ghost
 	for(var/mob/dead/observer/O in src.loc)
-		if(!O.client && !(jobban_isbanned(O, "catban") || jobban_isbanned(O, "cluwneban")))	continue
+		if(!O.client)	continue
 		if(O.mind && O.mind.current  && O.mind.current.stat != DEAD)	continue
+		if(!(jobban_isbanned(O, "catban") || jobban_isbanned(O, "cluwneban")))	continue
 		ghost = O
 		break
 	if(!ghost)
--- a/code/modules/research/xenobiology/xenobiology.dm
+++ b/code/modules/research/xenobiology/xenobiology.dm
@@ -403,7 +403,7 @@
 /obj/effect/golemrune/attack_hand(mob/living/user)
 	var/mob/dead/observer/ghost
 	for(var/mob/dead/observer/O in src.loc)
-		if(!O.client)	continue
+		if(!O.client && !jobban_isbanned(O, "catban"))	continue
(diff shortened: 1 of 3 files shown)


TheOneAndOnlyCreeperJoe:
Code's good.

ghost:
I don't know if this will runtime if a braindead catbanned ghost is on a rune. I'm gonna change the check to the next check when there is a client.

code/modules/research/xenobiology/xenobiology.dm
```diff
@@ -403,7 +403,7 @@
 /obj/effect/golemrune/attack_hand(mob/living/user)
 	var/mob/dead/observer/ghost
 	for(var/mob/dead/observer/O in src.loc)
-		if(!O.client)	continue
+		if(!O.client && !jobban_isbanned(O, "catban"))	continue
```
JohnGinnane: Two things:
1. This means that if the observer isn't a client AND they are not catbanned then it won't let them play as a golem. I take it this should be "OR they're catbanned"
2. Should we also exclude cluwnebanned people too?

ghost:
Cluwnebans probably should be added.
Shouldn't you be checking there is nothing that should make them fail. If one or both trues are false then it doesn't continue right?
 if(!O.client && !jobban_isbanned(O, "catban")) continue

ghost:
~~Did cluwnebans actually ever get fixed?~~
Apparently so, but I'm dumb and didn't know.

ghost:
Hold up I broke something.

ghost:
For some reason, the check only works in the first if statement?

ghost:
It was working then it broke somehow. This is sub-optimal.

trifornt-tinfornt:
Regardless, this is the best thing ever. You have no idea how obnoxious it is for some catbeast to show up as a Golem, run around spamming scream and promptly getting lynched by the other golems which is not only making me lose a golem but is annoying the hell out of me and wasting my time.

dreamcrab:
when this isn't broken, this is a very good idea

ghost:
> yfw CONTINUE doesn't act as continue on IF statement logic but as a CONTINUE TO THE NEXT ELEMENT of the LOOP

I think after a proper code review this should be good to merge after one more HA approval.

Nebyab:
someone code review this.

otherwise its fine

code/modules/research/xenobiology/xenobiology.dm
```diff
@@ -393,6 +393,7 @@
 	for(var/mob/dead/observer/O in src.loc)
 		if(!O.client)	continue
 		if(O.mind && O.mind.current && O.mind.current.stat != DEAD)	continue
+		if(!(jobban_isbanned(O, "catban") || jobban_isbanned(O, "cluwneban")))	continue
```
Kokojos: || is ''or'', so you're basicly doing :
If person has been EITHER not catbanned or not cluwnebanned, continue.

It should be, if either is true, then stop the thing.

if(jobban_isbanned(O, "catban")) ||  jobban_isbanned(O, "cluwneban") return

If jobanned from either, return null.

JohnGinnane: Remove the ! so you have the following:

`if (jobban_isbanned(O, "catban") || jobban_isbanned(I, "cluwneban")) continue`

This means if they're catbanned OR they're cluwnebanned then jump straight to the next iteration of the loop (and  do not execute anything after)

code/modules/research/xenobiology/xenobiology.dm
```diff
@@ -404,7 +405,8 @@
 	var/mob/dead/observer/ghost
 	for(var/mob/dead/observer/O in src.loc)
 		if(!O.client)	continue
-		if(O.mind && O.mind.current && O.mind.current.stat != DEAD)	continue
+		if(O.mind && O.mind.current  && O.mind.current.stat != DEAD)	continue
+		if(!(jobban_isbanned(O, "catban") || jobban_isbanned(O, "cluwneban")))	continue
```
JohnGinnane: See comment above. Same scenario here

ghost:
Literally about to do that, I think I had it working and forgot to commit it.

ghost:
Woowee, I fixed it.

Kokojos:
finnaly

PR #2417, opened 2016-05-23
Turns the labor camp into a secondary perma brig.

calibraptor:
Which is what it was all along really, considering it wasn't possible to redeem points and leave. This PR makes it much less ambiguous as to whether the labor camp is temporary or permanent.

Changes:
- Removed the box of prisoner IDs, since this change makes them useless.
- Removed prisoner management consoles, for the same reason as the prisoner IDs.
- Removed the non-functional point redemption consoles from the labor shuttle.
- Added a few chairs for bucklecuffing purposes.
- Added some lockers with prisoner jumpsuits to the labor dock.
- Removed the shutter that would block you off from actually mining stuff if it got closed, literally the only purpose of it was to piss off labor camp prisoners.

Majority of the functionality of the labor camp was left in, so patient officers can still bring perma prisoners to the labor camp and make them work for their freedom. The main purpose of this PR is to make it clear that the labor camp /IS/ an alternative perma brig, and thus shitty to throw people in there for petty crimes.


Nebyab:
Either have spesspermabrig or stationpermabrig. Permabrig but in spess is fairly pointless if we already have a permabrig stationside.

trifornt-tinfornt:
That and the labor camp is incredibly boring, honestly. I'd rather have the permabrig be some satellite out in space where you could still do things, but as of now it's either the one that's incredibly easy to break out of and the one that's boring and usually leads to the prisoners suiciding instead of actually mining and getting you metal.

PR #2427, opened 2016-05-25
Moves the pinpointer from Cap's Office to HoS Office

Sarcalogo:
And gives it to the HoS instead. It makes no sense for the Captain to start with the disk AND the pin pointer.

I map merged and this is the result. Created the branch from HippiestationCode so the map is up to date. Not sure if it conflicts with any other PR.


JohnGinnane:
Can you post screenshot where the disk is now?

JohnGinnane:
Also shouldn't the head of personnel get it seeing as he/she's next in line to the throne?

Sarcalogo:
Disk is in the exact same spot. Pinpointer is right where the universal recorder was on the HoS's desk. It's given to the HoS because he's the one who will be using it to track down the nuke disk in case of nuke ops or something.

JohnGinnane:
That does make more sense

Hackole:
I'm down for this. Not a bad idea at all

Crystalwarrior:
Definitely a good idea, otherwise nuke ops just toss the pinpointer and call it a day

McBawbaggings:
I'm inclined to disagree,
Captain should have the responsibility of delegating the second in command to guard the disk, it shouldn't be predetermined.

Sarcalogo:
It changes the location of the pin pointer, not the nuke disk. Right now, the captain starts with the nuke disk (incredibly important item) and the pinpointer (only way of tracking the incredibly important item). If the Captain dies, there is no way to track the disk unless the captain has handed of the pin pointer somewhere.

Gandalf2k15:
Makes little to no sense to move it, its like that everywhere. Just means some fag can steal it from the HoS's room, which has windows.

Sarcalogo:
It's not like that on /tg/, they changed it months ago. Ideally the HoS would carry the pinpointer with him at round start. And the windows are from space, or they'd have to break in through the entire brig.

trifornt-tinfornt:
This is stupid. The Captain is given the disk and the pinpointer for a reason. Why would the HoS get it when he doesn't have all-access or the disk?
The Captain should be the one responsible for the disk, and in case he somehow loses it, he can track it. The HoS wouldn't exactly need the pinpointer.

JohnGinnane:
I disagree trifornt, it makes sense if the disk is in one location, and the pin pointer is in another. It makes rushing Cap's office a little less effective because the pin pointer can still be used to locate the disk.

Kokojos:
I agree with this PR

Nebyab:
nah

McBawbaggings:
This is currently sitting at 2 headmin disapproval

PR #2429, opened 2016-05-25
Vampire Overhaul (WIP) and Shadowlings

Sarcalogo:
So the purpose of this overhaul may be a bit controversial, I'm not sure if it'll ever get finished because of that.

Basically, it seems like to me that every experienced player knows that shadowling is a broken game mode. Shadowlings to me seem like a fan fiction tier version of vampires. Vampires are kind of bare bones right now, and kind of lame. So uh.. the eventual goal of this PR is to get rid of Shadowling entirely and replace it with the vampire game mode. Some select shadowling powers will be given to the vampires.

Vampires ideally will work like they do on Goon and /vg/ with a mix of how they work on Hippie.

- Start off with a few powers
- Drain clean blood to gain more powers
- Try to make as many vampires as possible.
- Can't die unless they run out of blood or are gibbed or whatever.

So uh I'm thinking of giving the vampires something like: Lesser Glare
and nerfed shadow walk at the start.. haven't thought much out beyond that.

### DIFF
--- a/code/game/gamemodes/vampire/vampire.dm
+++ b/code/game/gamemodes/vampire/vampire.dm
@@ -72,6 +72,7 @@
 	V.verbs += /mob/living/carbon/human/proc/vampire_accelerated_recovery
 	V.verbs += /mob/living/carbon/human/proc/vampire_chiropteran_shapeshift
 	V.verbs += /mob/living/carbon/human/proc/vampire_demonic_strength
+	V.verbs += /obj/effect/proc_holder/spell/targeted/lesser_glare
 	V << "<span class='userdanger'>Lilith's blessing warps your body... you are a vampire!</span>"
 	V << "<b>You are a vampire - a supernatural, nearly-immortal creature nourished by blood. You are physically incapable of death unless certain conditions are met, and you will slowly recuperate all \
 	wounds while blood is in your body. However, there are some weaknesses that \ will bypass your immortality. Holy water and consecrated areas will either outright hurt you or at least weaken \


McBawbaggings:
>Lesser glare
Their suck incapacitates people and mutes them, they don't need the bullshit that is glare.

I'm not a big fan of vampires being a proper antag, but they're good midround antags imo.
>Shadow walk
Probably just needs reworking rather than be removed. If you outright remove it, shadowlings are gonna have an even rougher time.

TheOneAndOnlyCreeperJoe:
hurrdurr we don't like change.

I seriously don't like this. The current state of Vampires is that they stun people using basic shit, drain, rince and repeat until you get 500, then do the same thing all over again and convert them after draining them. This current state of affairs is dull but is fine in terms of balance. Adding proggresion to them wouldn't hurt.

But this? Lolno. Keep them as their current midround antag. Glare and Shadow Walk are the most stupid shit I've seen with Shadowlings and the entire community hates it. I do not need a carbon copy of shadowlings with a vampire cloak and pointy teeth.

Sarcalogo:
It's Lesser Glare which stuns for like 5 seconds. I also plan on removing the stun and mute when draining blood, so it's not so easy any more. Shadow walk I am nerfing right now in a seperate PR (my like 6th PR today lol). Plan is to double the cool down, and remove the stun removal on use. In other words, it will function just like etheral jaunt except with a doubled cool down.

TheOneAndOnlyCreeperJoe:
Have you even played vampire before? Draining blood's stun and mute is so incredibly sloppy, if you'd spam enough you can still talk because it repeatedly applies the effect within short bursts, so you're going to have to stun cuff anyway while abducting them, might as well take the headset while you're at it. In fact, you are making it easier to abduct people as a Vampire because the 5 second stun allows you either follow up with another stun or just cuff them on the spot, not to mention it allows you to fight back while stunned; which is a power only the Wizard should be entitled to!

Gandalf2k15:
Just fucking remove shadowling please.

Hackole:
agree with gandalf here, literally anything would be better than what shadowlings are now.

PR #2439, opened 2016-05-26
Fixes bluespace capsules fucking with the escape shuttle

blobbernaut:
they'll no longer create massive holes after the shuttle launches


Kokojos:
YES

carlarc:
a thing i've always wanted to do and never could was fill the shuttle with bluespace shelters and make it impossible to escape

---1

PR #2445, opened 2016-05-28
Fixes the permabrig posters

calibraptor:
Used the wrong object for the posters, this PR should fix it.


calibraptor:
Found a bunch more issues that need fixing, will be making a new PR for em

PR #2452, opened 2016-05-29
Minor Stimulant Changes

Kierany9:
Fixing a few things that went wrong in a 2-month old PR + some other minor balance changes, mostly to deal with powergaming cigar abuse while preventing the stupid but hilarious instant deaths that they caused.

- Removed a few useless REMs from the code I added by accident last time I touched narcotics because I didn't know what I was doing. This won't change anything as the value of REM is currently 1 but the things I added it on were not depenant on the metabolism rate such as stun durations and fixed ammounts of damage.
- Adds a cap to the maximum toxin damage you can take from Narcotics once they wear off. Crank is capped at 50, Meth is capped at 95 and Bath Salts is capped at 195.
- Added tolerance to Meth and Crank. Even if you regulate your doses to keep your drug levels. between 0 and 10u, their positive effects wear off in the time it would take you to overdose and you will quickly take toxin and stamina damage respectively. You'll know you've passed the tolerance limit when the high messages change.
- Adds scaling over time to Bath Salt's toxin damage, caps at 1 damage per tick after 45 ticks.
- Added a speed boost to Bath Salts and changed the messages you get while high.
- Added a delay on the positive effects of bath salts kicking in.
- Minor changes to synaptizine, has a chance to reduce stun length instead of halving their duration and being better than half of the narcotics plus easier to make and less toxic.
- Bath Salts now purges synaptizine. Sanity is not allowed.


McBawbaggings:
I don't think REM is ever useless.
If I go in the code and change the standard metabolism rate to 0.2 then all chems would leave the body twice as slow. If you don't use REM, then chems would be still working at the same rate which would double their performance.

McBawbaggings:
Wait no, I see what you mean now.

Kierany9:
Yeah, for anybody else who doesn't see it, it's for fixed stun durations in on_mob_delete that should not be affected by metabolism.

calibraptor:
+1

McBawbaggings:
More chems should be punishing for how long you have them rather than how much you have. +1

Sarcalogo:
Cool. This is necessary to merge so I can port

https://github.com/tgstation/tgstation/pull/17981
https://github.com/tgstation/tgstation/pull/17968

Nebyab:
okey dokey

Sarcalogo:
2 Heads said yes. Doesn't need head input as of 2 days ago...

McBawbaggings:
It needs the sweet code review though.

Carbonhell:
Code ok,merging

PR #2466, opened 2016-05-31
Byond 510 update

TDcoolguy300:
# BYOND 510 compatibility

As above, this should allow us to run 510 without too many issues, in my tests everything was running just fine, ~~with the exception of the on-screen flashes looking a tad odd~~ Fixed it

Big thanks to the coders over at TG with testing this stuff before us, namely Neersighted and MrStonedOne


calibraptor:
Aw shiiiiiit

TDcoolguy300:
At this point, only Nalar or Hippie may merge this, because this will break the server should we not update BYOND

nalarmg:
Do not merge under any circumstances for the time being. I'd like to wait a little and make sure it's safe before we upgrade.

nalarmg:
This hasn't been forgotten. Currently needs extensive testing which is only available via the main server. Details on how to make a test round work with this are being worked out.

PR #2472, opened 2016-06-03
Improve travis script error checker

optimumtact:
Builds should not pass when they have an error count that is a multiple of 10

basically a dreammaker output of 
- 10 errors and 0 warnings was matching the grep  like so : - 1[0 errors and 0 warnings]
  the new grep doesn't match, as it looks for the - as well, so it will now only match
- 0 errors and 0 warnings.

### DIFF
--- a/dm.sh
+++ b/dm.sh
@@ -79,7 +79,7 @@ else
 	then
 		DreamMaker $dmepath.mdme 2>&1 | tee result.log
 		retval=$?
-		if ! grep '0 errors, 0 warnings' result.log
+		if ! grep '\- 0 errors, 0 warnings' result.log
 		then
 			retval=1 #hard fail, due to warnings or errors
 		fi


danhibiki:
This is a fix that is live on TG and has also been submitted by the same person to VG and bay, they likely know what they are talking about.

optimumtact:
@nalarmg

Carbonhell:
Already asking him, i assume the \ - operator is used to strict check a text string,correct?I'm asking him to be sure before merging.

optimumtact:
@Carbonhell in this case it's doing a match against the - in the output from dream maker

dreammaker outputs the literal string

```
{some other stuff }  - 0 errors and 0 warnings
```

on a successful build, so by including that in the grep as the literal character -, I eliminate the possibility of matching against

```
 {some other stuff } - 10 errors and 0 warnings
```

Which is of course obviously not what we wanted

Carbonhell:
Understood, merging. Thank you man!

PR #2489, opened 2016-06-06
Z-Level 5 Fixes

calibraptor:
- Fixed the permabrig shutters so they are properly layered with the doors.
- Placed a plant bag, plant analyzer, water tank, bucket, and dull hatchet (6 force) to the permabrig recreation area so people can do proper botany there.
- Added a shuttle console to the northern mining outpost so scientists don't get stranded there. Also added proper lighting to the hallway.
- Moved poorly placed fire extinguisher cabinet to a more sensible spot.


Spacedong:
does the dull hatchet have a sprite like the dull knife/rolling pin

TDcoolguy300:
Map fixes, mergin

Issue #2382, opened 2016-05-20
Infected monkeys with facehuggers do not burst


darklizard45:
When you infect a monkey with a facehugger it doesn't burst no matter how long you wait, it also doesn't shows any symptoms

Gandalf2k15:
I'll take a bash at this.

Gandalf2k15:
Fixed. https://github.com/HippieStationCode/HippieStation13/pull/2494

PR #2494, opened 2016-06-06
Monkey impregnation bugfix

Gandalf2k15:
Basically there was a retarded check that was not needed and it caused monkey's to fail it.
Fixes https://github.com/HippieStationCode/HippieStation13/issues/2382
Impregnated:
image
Burstie:
image


JohnGinnane:
Code looks good, merging as this is a bug fix.

PR #2511, opened 2016-06-09
Firing pin additions

Hackole:
Adds a syndicate firing pin, which will deny firing whilst at the syndicate outpost. This will hopefully cut down on nuke-teams blowing themselves up before they even get to the station. The pin has been added to most guns that they can get their hands on.

Adds a trick firing pin to uplinks for 3tc. Upon trying to fire a weapon with this pin installed, it will explode, likely critting whoever the victim is.


trifornt-tinfornt:
http://i.imgur.com/xcjS4MP.png
it sucks
On a more serious note, it's a godsend. No more idiotic leaders declaring war and buying blastco without asking their team which always leads to the inevitable newbie Op misfiring the Striders and killing everyone.
Trick Pins sound like fun, though they should have tacticool pin replacement where if you simply place it in a gun, pin or not, it'll automatically place its self in instead.

myontasm:
came in here expecting shitty firing pin additions (as the whole firing pin thing is unused and thus useless)

this is very good I like it

ghost:
Good shit +1

ArcLumin:
Excellent. I like it

Gandalf2k15:
+19999

dreamcrab:
this is creative and cool +420

Voratius:
Sure, sounds like a neato featuro +1/3rd.

Have you playtested it completely to make sure there aren't any issues?

Carbonhell:
god no,the idea behind it is ok but the way it's implemented isn't really good,i'd have preferred if it checked the area it's in, aka make an area setting that disallows any kind of fight or idk

code/modules/projectiles/guns/projectile/revolver.dm
```diff
@@ -102,6 +103,7 @@
 	mag_type = /obj/item/ammo_box/magazine/internal/cylinder/rev38
 	unique_rename = 1
 	unique_reskin = 1
+	pin = /obj/item/device/firing_pin/generic
```
JohnGinnane, line 106: How come these guns get generic pins?

code/modules/projectiles/pins.dm
```diff
@@ -112,6 +112,32 @@
...
+/obj/item/device/firing_pin/area/syndicate
+	force_replace = 1
+	A = /area/syndicate_mothership
+	name = "Syndicate firing pin"
+	desc = "This pin will prevent firing whilst at the operative outpost."
+	icon_state = "firing_pin_pindi"
+
+/obj/item/device/firing_pin/trick
+	name = "Generic firing pin"
+	desc = "This pin will detonate the weapon it is put into upon trying to use it"
```
JohnGinnane, line 135: > ..detonate the weapon it is put into upon trying..

I'm not sure I understand what this is trying to say?

JohnGinnane:
I kinda disagree Carbon. This is pretty straightforward and doesn't require changing areas. It also means you could add a pin to stop some mining equipment being used outside the mining area potentially, again without having to change area definitions too much

Spacedong:
pretty good and I can see it being used for other stuff too maybe in the future, just make sure grenade launchers have them, thats the most important thing

code/modules/projectiles/pins.dm
Hackole, line 135: It's saying it detonates the weapon when it is fired with the pin in - bad wording on my part.

code/modules/projectiles/guns/projectile/revolver.dm
Hackole, line 106: The parent revolver was the syndicate variety available to nukeops, so I had to edit it to have the syndicate pin. It wouldn't make sense for the other ones to have syndicate pins too, so I just made them have generic ones.

Hackole:
Yes spacedong, I ensured that the grenade launchers had the pins. That was the biggest offender.

Kokojos:
This has reached the votes, and code seems fine.

PR #2524, opened 2016-06-13
R&D bugfix

Gandalf2k15:
Basically fixes one error in weapons_designs.dm. Incorrect file path and incorrect mat amount.
E.G it was a copy of pistol.


JohnGinnane:
What was the error? As far as I can see that buid_path is valid: https://github.com/HippieStationCode/HippieStation13/blob/652a98380859564d3c483761c2b095a2501ace82/code/modules/projectiles/guns/energy/laser.dm#L24

JohnGinnane:
It looks good but I'm surprised to see an error

Gandalf2k15:
It prints out as a pistol instead of a rifle and has the same values.

Gandalf2k15:
I never noticed it in development. So people will try to create it and it will come out as a rifle. This does not need head input nor is it balancing.

JohnGinnane:
Oh whoops, I see now. My bad. Merging as it's a bug fix

Issue #2371, opened 2016-05-18
Stunprod charge bug


Kierany9:
Stunprods made with normal powercells show up with 100% charge when inspected but say that the cell is empty when turned on. IIRC stunprods are only supposed to be useable with higher quality powercells.

McBawbaggings:
Stunprods should only be high capacity+.

Kierany9:
Yes but you can craft one with a normal cell(which is the problem) and upon inspection it will say it's fully charged but when turning it on it says it's empty. You can't put the cell back in if you take it out.

McBawbaggings:
Did you table craft it?

Kierany9:
Yes, IIRC.

Gandalf2k15:
Unable to recreate.
http://i.imgur.com/bekc6ZO.png

Kierany9:
Replicated here, am currently fixing

https://www.dropbox.com/s/o9h0bxs97vbx8tq/Desktop%2006.18.2016%20-%2013.11.17.02.mp4

TDcoolguy300:
Whoops

Issue #1963, opened 2016-03-06
Deconstructing and constructing nukeop shuttle terminal fucks it up


Carbonhell:
it's due to the computer itself using the same board of normal shuttles, it needs a dedicated board,easy fix

PR #2542, opened 2016-06-18
Bugfixes

Kierany9:
Fixes https://github.com/HippieStationCode/HippieStation13/issues/2371 and https://github.com/HippieStationCode/HippieStation13/issues/1963

 - Nuke Ops shuttle consoles no longer break after being deconstructed.
 - Stunprods are now crafted without powercells, you'll need to find and add one of high enough capacity for it to work of course.


JohnGinnane:
Something tells me the stunprod fix won't fix it. The issue stems from the requirement being `/obj/item/weapon/stock_parts/cell` as seen here: https://github.com/HippieStationCode/HippieStation13/blob/master/code/modules/crafting/recipes.dm#L48

Kierany9:
Power cell is now added manually to tablecrafted stunprods to avoid having to add a snowflake var check into the crafting proc and because Bawb doesn't want stunprods to be craftable with power cells that are too weak.

JohnGinnane:
Code looks good. Did you test deconstructing the shuttle consoles and also manually and table crafting stunprods?

Kierany9:
Yes

code/modules/crafting/recipes.dm
```diff
@@ -44,10 +44,8 @@
 	result = /obj/item/weapon/melee/baton/cattleprod
 	reqs = list(/obj/item/weapon/restraints/handcuffs/cable = 1,
```
Carbonhell: Won't this give a prod with a cell in it

Carbonhell: I mean,the result var

html/changelogs/Kierany9 - Bugfixes
```diff
@@ -0,0 +1,5 @@
+author: Kierany9
```
Carbonhell: You forgot the .yml extension

html/changelogs/Kierany9 - Bugfixes
```diff
@@ -0,0 +1,5 @@
...
+delete-after: True
+changes: 
+- bugfix: "Nuke Ops shuttle consoles no longer break after being deconstructed."
```
Carbonhell: And the proper indentation here too

code/modules/crafting/recipes.dm
Kierany9: No, the cell was inserted in the parts list.

PR #2546, opened 2016-06-20
Fixes the see-through-walls bug

TDcoolguy300:
Potentially fixes the 'I CAN SEE THE DISK THROUGH THE WALL!' bug, worked in my test build. http://puu.sh/pzlrV/abab3c0565.png


Carbonhell:
You may need to escape the 's in the changelog,unsure

TDcoolguy300:
We'll see when Travis complains

Issue #2535, opened 2016-06-15
Cant refine carp fillets into regents


Erawh:
Carp fillets cant be grinded in a all-in-one grinder. 
Place carp meat into a grinder to test

GenericGoose:
I don't yet know how it all works (or rather doesn't) but basically all snacks aren't grindable
https://github.com/HippieStationCode/HippieStation13/blob/5462010e5bcc8a61ad5618ffd0f04f07eeb570b0/code/modules/reagents/Chemistry-Machinery.dm#L1003
maybe this check doesn't work
https://github.com/HippieStationCode/HippieStation13/blob/5462010e5bcc8a61ad5618ffd0f04f07eeb570b0/code/modules/reagents/Chemistry-Machinery.dm#L1099

I totally remember that I was able to grind food into nutriments and meat was grindable sometime ago too.
Don't yet know how to fiks it. Also gotta refactor chemistry_machines, 1.5k lines of code and a lot of random machines, impossible to track history

Another theory is there's comment saying 'put all recipes above this' and then a recipe right below...If that's the case we can blame [this commit ](https://github.com/HippieStationCode/HippieStation13/commit/e9f5e9628911b0c5019cbca2f3e71e330fe9b4bb)
Nah, that's not it

GenericGoose:
Alright[ this commit](https://github.com/HippieStationCode/HippieStation13/commit/7f33129131768f0fa5538787d5d09ae24a535c12) is to blame, I guess.
I'll revert it and check if things work. If they do i'll PR a revert.
Yeah, it really broke the grinder, I reverted and it works.

PR #2522, opened 2016-06-12
CQC Gloves Disguise

GenericGoose:
Allows you to disguise CQC gloves as any other gloves (not including colored), so they aren't useless to jobs that have their own gloves or for sneaky fun (works same way chameleon thermals do, btw someone should probably re-write that proc to make it universal, which may be the nekst thing i'll do). Only changes name and sprite.

Upd: now this PR also fikses [missing sec bomb suit icon](https://github.com/HippieStationCode/HippieStation13/issues/2530), changes is_type_in_list back to what it was before 'optimization' that broke it, so it [fikses ](https://github.com/HippieStationCode/HippieStation13/issues/2501)reagent grinder [not being able to process any food](https://github.com/HippieStationCode/HippieStation13/issues/2535) and replaces Pete's attack message for some reason it bothered me. Oh yeah and I added disguising feature to pickpocket gloves too as @Kierany9 suggested

PR #2551, opened 2016-06-20
Fixes grinder - CREDITS: GeneralGoose

Gandalf2k15:
Fixes grinder, see https://github.com/HippieStationCode/HippieStation13/pull/2522/files.
Fixes https://github.com/HippieStationCode/HippieStation13/issues/2535

### DIFF
--- a/code/__HELPERS/lists.dm
+++ b/code/__HELPERS/lists.dm
@@ -50,8 +50,8 @@
 	if(!L.len)
 		return 1
 	return 0
-
-/proc/is_type_in_list(atom/A, list/L) //Thank RemieRichards for writing this
+//Checks for specific types in a list
+/proc/is_type_in_list(atom/A, list/L)
 	for(var/type in L)
 		if(istype(A, type))
 			return 1
--- a/code/__HELPERS/lists.dm
+++ b/code/__HELPERS/lists.dm
@@ -51,7 +51,6 @@
 		return 1
 	return 0
 
-//Checks for specific types in a list
 /proc/is_type_in_list(atom/A, list/L) //Thank RemieRichards for writing this
 	for(var/type in L)
 		if(istype(A, type))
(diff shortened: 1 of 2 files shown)


blobbernaut:
This is a bad way to fix grinders as it reverts a proc that was made literally dozens of times faster and is used everywhere

Gandalf2k15:
it works though.

JohnGinnane:
Can you try to fix it without reverting the PR though?

GenericGoose:
i just found it and suggested to revert it.

blobbernaut:
@RemieRichards help

RemieRichards:
@blobbernaut I actually did the same fix here, except I kept mine and MSO's faster version as a specialist proc (with another proc to assist in building typecaches) and applied it in places that really needed it, like the Tesla, Hostile AI, etc.

GenericGoose:
yeah today i got same idea, to leave new version and re-add old and apply to the broken thing
but really someone's gotta make the new one work

GeneralGoose:
http://puu.sh/pB0tF/dd4686f51d.png
selected tile has no disposal pipe
gg
not like anyone uses security criminal disposals anyway (because it wasnt reported)

ghost:
We should probably merge this after code freeze at the very least.
Sure we won't have the efficient version, but it beats having broken things doesn't it?
We can always refactor the proc back in like Remie has suggested.

code/__HELPERS/lists.dm
```diff
@@ -53,11 +53,10 @@
 
 //Checks for specific types in a list
 /proc/is_type_in_list(atom/A, list/L) //Thank RemieRichards for writing this
```
JohnGinnane: I might be nitpicking, but would you mind removing the comment too?

JohnGinnane:
Code looks good. I will merge if comment is remove unless you think it's irrelevant

JohnGinnane:
Code looks good. Going to merge as this is (hopefully) a temporary bug fix. The plan is to implement this again but use the faster proc for heavier things instead, as suggested by Remie

PR #2555, opened 2016-06-21
Possibly fixes the Reagent Metabolization issue

TDcoolguy300:
Somehow datums are not being properly removed from the obj SubSystem. This method is no different than what i replaced, but if it works, ¯_(ツ)_/¯. Unable to reproduce the runtimes locally however.


Carbonhell:
No this won't fix it man,pretty sure

GeneralGoose:
what was the issue

TDcoolguy300:
Then i have no idea Carbon, Reagents datums are either not being removed from the obj subsystem on deletion, or are straight up having read issues.

Nebyab:
will it work or not, who knows?

Carbonhell:
Why the hell didn't i close this pr?
im dumb

Issue #2543, opened 2016-06-19
Dismembered heads don't seem to be affected by conveyor belts or the mass driver.


TrashFiend:
I took a head I dismembered and dropped it on the conveyor belts in disposals and it wouldn't move. I then just manually moved it onto the mass driver. While everything else fired off into space, the head just stayed there. Seems weird.

TDcoolguy300:
This is because they are "anchored", they can not be moved by normal means such as dragging, belts, and are too massive to fit in crates. But you can still pick them up because they are still items.

Kokojos:
Simple 1 var change fix I mean, why would they ever need to be anchored?

ghost:
It's so you cant drag them.

Same reason you cant put them in a backpack and it slows you to carry them, they're supposed to be hard to transport so you dont get fuckboys stealing heads I guess.

trifornt-tinfornt:
I have no idea why these would not only need to be gigantic, but also be anchored. It's not like you can't just remove the brains from the head and carry on.

Carbonhell:
That takes time,whereas chopping someone's head off and running away With it is Too op

trifornt-tinfornt:
Not really, it takes a matter of seconds - and any sharp object can remove the brain so once you dismember them it's another two seconds to take their brain out. Not overpowered at all.

GenericGoose:
you can drag bodies but can't drag heads
whats the fucking point it takes longer to chop off a head
you still can put them in a locker in maint and nobody will ever find them

Kokojos:
Technicly not a bug, shitty feature that's apparently here to stay.

PR #2563, opened 2016-06-23
Fixes heads

Gandalf2k15:
Heads are no longer retarded. They can be pulled and put in bags.
Seriously who thought this was a good idea?
fixes https://github.com/HippieStationCode/HippieStation13/issues/2543

### DIFF
--- a/code/modules/surgery/organs/organ_external.dm
+++ b/code/modules/surgery/organs/organ_external.dm
@@ -194,15 +194,12 @@
 	Plenty of additional icon appearance bullshit in here, watch out.
 */
 
-/obj/item/organ/limb/head
+/obj/item/organ/limb/head //Crystal why did you do this.
 	name = "head"
 	icon_state = "head"
 	max_damage = 200
 	body_part = HEAD
-	w_class = 4 //Quite a hefty load
-	anchored = 1 //Forces people to carry it by hand, no pulling!
-	slowdown = 1 //Balancing measure
-	throw_range = 2 //No head bowling :(
+	w_class = 2
 
 	px_x = 0
 	px_y = -8
--- a/html/changelogs/Gandalf2k15 - Head fix.yml
+++ b/html/changelogs/Gandalf2k15 - Head fix.yml
@@ -1,0 +1,36 @@
+################################
+# Example Changelog File
+#
+# Note: This file, and files beginning with ".", and files that don't end in ".yml" will not be read. If you change this file, you will look really dumb.
+#
+# Your changelog will be merged with a master changelog. (New stuff added only, and only on the date entry for the day it was merged.)
+# When it is, any changes listed below will disappear.
+#
+# Valid Prefixes: 
+#   bugfix
+#   wip (For works in progress)
+#   tweak
+#   soundadd
+#   sounddel
+#   rscadd (general adding of nice things)
+#   rscdel (general deleting of nice things)
+#   imageadd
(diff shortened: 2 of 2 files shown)


JohnGinnane:
Code looks fine, I wonder if anything relied on the heads being like this. Did you test messing around dismembered heads

Gandalf2k15:
Yep

JohnGinnane:
Do we require head input on this?

Gandalf2k15:
Good question.

ghost:
I'd rather have heads still be bulky and not to be able to remove their brain by hitting it.
That alone is really garbage.

Gandalf2k15:
but it makes no sense

ghost:
So?
Decapping people and hiding their heads pretty much is gibbing them. There should be at least some sort of difficulty associated with it. Decapping and cutting the brain out is easy as fuck and does the same thing. It bypasses an intended gameplay balance choice because memes. Making it so the bypass is the norm doesn't solve the problem of decapping being garbage for the victim.

Gandalf2k15:
its easier to hide a body than it is a head cause you can drag them. At least make it dragable/throwable

ghost:
You can't fit a corpse in a backpack is the problem as I see it, and debraining a corpse takes tools and/or time. Also heads are super easy to hide.

Personally I'd make them not retarded in how they move, but keep them too big for backpacks.

ghost:
Unless you full strip then no due to SUIT SENSORS. Dragging bodies usually leaves a blood trail too. Killing someone, decapping them and taking their brain ensures they will never ever be cloned. Debraining from decapping takes no time at all.

GenericGoose:
implying people follow blood trails
and chances are, if you are able to decap someone you won't be caring about someone trying to clone whoever you killed (because in my eksperience its just unstoppable murderbone from which there is no possible recovery)

Nebyab:
make heads bulky

Kokojos:
Bulky heads are stupid.
+1/3 to this

Carbonhell:
Heads please discuss this
@McBawbaggings 
@Nebyab 
@teamstabwhateveryouare

Spacedong:
@johnginnane

JohnGinnane:
what

McHatters:
Heads should be able to fit in bags if brains can.

dreamcrab:
i like heads being something bulky, make it so you have to debrain them to be more stealthy

Carbonhell:
needs news in 2 days or pr will be closed

JohnGinnane:
Heads can you give your input please

ghost:
To be fair, I'd rather have heads remain bulky and no longer have the mechanic of slicing a head open to extract brains.

Stealing heads is a sure fire way to make a body uncloneable and it's not that hard to do. Debraining takes a tad bit longer but the same applies.

blobbernaut:
Tbh, I'd rather bodies be cloneable without the head especially since heads are small, hard to find, suffer occasional sprite issues making them invisible, and also really easy to remove, and can't have suit sensors.

It's almost as good as gibbing and I don't like it, cloning should first look for an attached head+brain, then look for a heart if there's no brain.

blobbernaut:
Also I'd like heads be able to be put in MMI's.

JohnGinnane:
I agree, you should be able to clone using decapitated body or head (or both), and then we could treat heads like we do brains

PR #2575, opened 2016-06-27
Fixes AIs being perma blind.

ghost:
The clear_fullscreen only occured if you fixed the power shortage before 5 seconds.
This adds it to the other cases where power is restored after those 5 seconds.


TDcoolguy300:
This is a bug fix, merging

PR #2585, opened 2016-07-06
Mitigates drone shittiness

ghost:
Drones can no longer:
- Attack non-drones
- Help non-drones
- Shoot guns
- Blow up welder tanks

Syndie drones remain untouched.

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -25,6 +25,13 @@
 	I.attack(src, user)
 
 /mob/living/proc/attacked_by(obj/item/I, mob/living/user, def_zone)
+
+	if(isdrone(user))
+		var/mob/living/simple_animal/drone/D = user
+		if(!D.can_interfere(src))
+			D << "<span class='danger'>Your laws prevent you from doing this!</span>"
+			return
+
 	var/dmgcheck = apply_damage(I.force, I.damtype, def_zone)
 	if(!dmgcheck)
 		visible_message("<span class='danger'>[user] has attempted to attack [src] with [I].</span>",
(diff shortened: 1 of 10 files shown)


ghost:
I should mention some of this code is working around limitations that I don't have the knowledge to refactor (such as pockets/embedded items). If anyone can think of a more elegant way to implement it then please lemme know.

ArcLumin:
Goddamn it, was thinking we'd have Midgetgates

I like this shit, looks good.

Nebyab:
ok

code/_onclick/item_attack.dm
```diff
@@ -25,6 +25,13 @@
 	I.attack(src, user)
 
 /mob/living/proc/attacked_by(obj/item/I, mob/living/user, def_zone)
+
+	if(isdrone(user))
```
Carbonhell, line 29: This check is pretty snowflakey here,it would be better if it was on the drone's side,somehow

code/game/objects/items/weapons/tools.dm
```diff
@@ -328,6 +328,11 @@
 				user << "<span class='notice'>[D] has not enough welding fuel to refill!</span>"
 				return
 		else
+			if(isdrone(user))
```
Carbonhell, line 331: eh

Carbonhell:
the idea is good but it's the most snowflakey pr i've ever seen
i can't ok it anyhow,i am sorry

ghost:
I understand, from what I can see there is no proc for a drone attacking things but rather things being attacked by drones.
I checked /vg/ code which used mommi which uses a similar technique as what I did.
The other plan is to go over drone Onclick code and override that.

ghost:
The welder tank proc could probably be moved into the reagentfueltank object and such.
I also didn't realise the gun proc has a can_fire proc which I could piggyback on.

ghost:
Remaking this.

PR #2588, opened 2016-07-08
Revert "Fixes grinder - CREDITS: GeneralGoose"

ArcLumin:
Reverts HippieStationCode/HippieStation13#2551

Numerous issues abound with this, which just hit. Can't grind plasteel, food still ungrindable, multiple things have become ungrindable, even more than before.


ArcLumin:
Confirmed, only 3 things grind now. Plasma, Corn, and Arabica. Literally, the first on every list. Fucks up microwaves too

ArcLumin:
99% sure this is also why suit storage doesn't work now, holy shit.
https://github.com/HippieStationCode/HippieStation13/issues/2589

Likely the lists change now makes it so that only the first item in ANY list counts.

Gandalf2k15:
Merging as its a bugfix, code looks good. Also its a revert so things should grind again!

PR #2595, opened 2016-07-09
fixes a priority supermatter bug

### LINKED ISSUE #2567: Sticking a bunch of people/bodies in an object then slamming it against a supermatter will kick all players involve and let them respawn
Longass title.

Basically, I first noticed this in a ripley, holding a closet of 3 dead bodies. Went to go lift the supermatter into my cargo compartment when lagspikes hit and I ran into the supermatter. Immediate kick and disconnect from the game, when I logged back in, I was able to rejoin as a fresh new character, and not just me, but everyone who was a dead/ghosted that I had in the locker was kicked and forced to make a new character

Steps to replicate:
Load a locker of players into a ripley
Walk into the supermatter

Spacedong:
- fixes #2567 in which people would get kicked in a mech or closet when bumped into the shard

- replaced an instance of del with qdel

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -336,7 +336,7 @@
 
 
 /obj/machinery/power/supermatter/proc/Consume(atom/AM as mob|obj)
-	if(istype(AM,mob/living))
+	if(istype(AM,/mob/living))
 		var/mob/living/user = AM
 		user.dust()
 		power += 200
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -324,9 +324,6 @@
 
 
 /obj/machinery/power/supermatter/Bumped(atom/AM as mob|obj)
-	if(istype(AM,/obj/mecha) || istype(AM,/obj/structure/closet))
-		qdel(AM)
-		return
 	if(istype(AM, /mob/living))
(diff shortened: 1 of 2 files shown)


code/modules/power/supermatter/supermatter.dm
```diff
@@ -324,6 +324,9 @@
 
 
 /obj/machinery/power/supermatter/Bumped(atom/AM as mob|obj)
+	if(istype(AM,/obj/mecha) || istype(AM,/obj/structure/closet))
+		qdel(AM)
```
JohnGinnane: What does qdelling a closest do to the contents? I see in https://github.com/HippieStationCode/HippieStation13/blob/5462010e5bcc8a61ad5618ffd0f04f07eeb570b0/code/game/objects/structures/crates_lockers/closets.dm#L191 it first empties the contents and then qdels it. Should we do the same here?

JohnGinnane: To add to that is there a proc that handles qdel?

TDcoolguy300: Should an item have it's own defined Destroy, Qdel will call that before deleting the object, otherwise it's deleted safely.

ArcLumin:
It's a pretty gamebreaking, albeit fairly rare bug. Looks like a good fix.

Gandalf2k15:
Code looks good, merging as its a bug fix

Issue #2592, opened 2016-07-09
Sm does not create 13x13 sing


Gandalf2k15:
WHERE IS MY SUPERMATTER CHARGED SINGULO

ArcLumin:
If I recall, SM going into Sing unlocks it's 4th tier which it'll grow to once it absorbs enough? I don't think it goes 13x13 off the bat, but goes 13x13 on it's 4th tier once it eats enough

PR #2600, opened 2016-07-10
Multiple High Priority Supermatter Fixes

Gandalf2k15:
Fixes multiple game breaking issues with the supermatter, they are as follows.
1. The supermatter used Del() instead of Destroy(), addressed.
2. The supermatter used del to remove itself upon explosion, replaced with qdel in conjunction with Destroy().
3. The suppermatter never logged or created a noise when it consumed someone or something.
4. The supermatter never irradiated properly when it was bumped into, now uses radiation_pulse.
5. The supermatter SILENTLY created a stage six singularity, with no log or sound. IT now does.
6. On consumption it never logged, now it does along with the fix for consuming mechs and containers.

Fixes https://github.com/HippieStationCode/HippieStation13/issues/2592

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -95,8 +95,8 @@
 	radio = new (src)
 
 
-/obj/machinery/power/supermatter/Del()
-	del radio
+/obj/machinery/power/supermatter/Destroy()
+	qdel(radio)
 	. = ..()
 
 /obj/machinery/power/supermatter/proc/explode()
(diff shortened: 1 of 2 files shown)


Nebyab:
ok

code/modules/power/supermatter/supermatter.dm
```diff
@@ -307,44 +311,58 @@
 			R.receive_pulse(power * POWER_FACTOR * (min(3/distance, 1))**2)
 	return
 
-/obj/machinery/power/supermatter/attackby(obj/item/W as obj, mob/living/user as mob)
+/obj/machinery/power/supermatter/attackby(obj/item/W, mob/living/user, params)
```
JohnGinnane, line 314: What's with params?

Gandalf2k15, line 314: Its to determine if its the Shard or the SM

JohnGinnane:
Looks good. Merging as it's priority

PR #2553, opened 2016-06-20
Admin Ticket System Ver. 2

Gandalf2k15:
Basically this is a more robust version of the one I implemented some time ago. This has been moulded to the feedback I received.
It enables admins to track ahelps and when they join they will see any unresolved ahelps. Its all text based apart from viewing logs which is a window that pops up. Players and admins can view ahelp logs. Makes shit easier. Please gives me feedback and what you want and shit.
This means that you dont NEED to use it. You can stick to the regular shit if you want.

Steps are easy, fairly self explanatory. Adds a new Admin Help tab for ease. TEST IT TO SEE

Everything is good and done.

PR #2616, opened 2016-07-12
Revert "Admin Ticket System Ver. 2"

Spacedong:
Reverts HippieStationCode/HippieStation13#2553

:ok_hand:


Hackole:
If you don't want to use it, then don't. Really the only thing this changes for an admin that isn't using it is that it gives a popup telling you if someone is handling an ahelp, preventing pileups on a single ahelp which sometimes happens.

BroomHill:
Please revert this shit. I know it doesn't change much but the whole thingy messes with me.

Gandalf2k15:
Again, don't have to use it with the update. Dunno how it could possibly mess with you.

PR #2652, opened 2016-07-21
Removes bluespace shelter capsules

ghost:
Removes them entirely.
I'll reiterate what I've said previously:
"I don't think they're fit for purpose.
The whole point of the escape pods being used for emergency escape during red delta is for people who want to run away during malf/nukies/gangs etc.
You get mining gear and spacesuits to escape the asteroid, you don't need a little autism fort on top."

They have a bunch of snowflakey bugs associated with them, they go against the whole escape from the station with emergency gear and are not required.


ghost:
Oh fuck.

PR #2658, opened 2016-07-22
Inflatable Barrier Revamp - CREDITS: Carbonhell

Gandalf2k15:
Walls are now revamped, with good code.
They will now allow air past when destroyed/deleted.
Credits to Carbonhell for revamping it.
No balancing was changed, therefore its a bugfix.


Carbonhell:
code is obviously ok,merging

Issue #2668, opened 2016-07-24
You can lock open lockers


ArcLumin:
Unsure if this is a bug but it's a bit of a balancing issue. You can open a lockable locker, drag something on it to prevent it from closing and click on it, it'll lock instead of closing. Then you can have it both opened and locked, and have it be locked whenever it closes

This allows you to easily get anyone to stand on the open locker tile in the halls, close it and they can't escape.

ghost:
The mere fact you can instantly capture people like this is pretty funny though

blobbernaut:
This is great, don't remove

ghost:
Not the right attitude tho

PR #2676, opened 2016-07-25
Fixes being able to lock open closets.

ghost:
Fixes https://github.com/HippieStationCode/HippieStation13/issues/2668


Gandalf2k15:
Code's fine. But how could you this would be top prank!

PR #2689, opened 2016-07-30
Adds stolen Colonial Marines code to the server

Kierany9:
honk

- Replaces the Gooncode disk with a Colonial Marines Code disk.


McBawbaggings:
>replaces goon disk

That alone is questionable.
If CM's staff want their code to be kept secret then it needs to self destruct on deconstruction.

Pyko1:
Why remove the gooncode?

Could keep it and add this as a new joke instead.

Badmex:
This should not remove the gooncode. This should just add it as another disk.
-1/3 for replacing the goon disk. You get a +1/3 if you add it as extra!

code/modules/research/research.dm
```diff
@@ -323,3 +323,11 @@ research holder datum.
 	icon = 'icons/obj/items.dmi'
 	icon_state = "gooncode"
 	stored = new/datum/tech/programming{level = 4}
+	
+/obj/item/weapon/disk/tech_disk/gooncode
```
Carbonhell: change the path ending to cmcode

code/modules/research/research.dm
```diff
@@ -323,3 +323,11 @@ research holder datum.
...
+	name = "Colonial Marines Code disk"
+	desc = "A disk containing Colonial Marines source code for further research."
+	icon = 'icons/obj/items.dmi'
+	icon_state = "cmcode"
+	embed_chance = 85
```
Carbonhell: wut,why

Pyko1:
I think 4 might be a bit too much even if its a pain in the ass to get.

Kierany9:
[internal screaming]

McBawbaggings:
[CM CODE WAS EXPUNGED AS BASED ON THE TERMS]

GeneralGoose:
failure

PR #2691, opened 2016-07-30
Medical Cyborgs now have Tricordazine instead of the 4 main healing chems. Emagged mediborgs now get CLF3 instead of Facid.

ghost:
You now cycle through chems as well instead of the menu selection.
This is technically a nerf since tricord works slower than the other chems.
But I'd really did prefer the old healing system where it was fire and forget.

NEW THING:
This now changes emagged mediborgs to use CLF3 to make them a far more viable option for traitors and such. Ever since acid was made to tickle and we brought back sec borgs, mediborgs have been cast away in terms of usefulness.

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -159,15 +159,15 @@
 	add_module(new /obj/item/stack/medical/gauze/cyborg())
 
 	emag = new /obj/item/weapon/reagent_containers/spray(src)
-	emag.reagents.add_reagent("facid", 250)
-	emag.name = "Fluacid spray"
+	emag.reagents.add_reagent("clf3", 250)
+	emag.name = "chlorine trifluoride spray"
 
 	fix_modules()
 
 /obj/item/weapon/robot_module/medical/respawn_consumable(mob/living/silicon/robot/R, coeff = 1)
 	..()
 	if(R.emagged && istype(emag, /obj/item/weapon/reagent_containers/spray))
-		emag.reagents.add_reagent("facid", 2 * coeff)
+		emag.reagents.add_reagent("clf3", 2 * coeff)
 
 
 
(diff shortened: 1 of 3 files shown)


calibraptor:
Yes please +1/3rd

Kierany9:
Why not have both?

ghost:
Because cycling through all the better chems would be awful.
And the menu was an awful idea in the first place.

Carbonhell:
Sidenote,DD drains nutriton if taken by nonmedics

Badmex:
Yes +1/3

ghost:
Intended kinda, it's either use tricord which is even weaker than current DD, has an OD limit.
I'd rather have old DD's but I think that would fuck balance to hard in the mess we call our chemistry system.

Carbonhell:
Code is ok

ghost:
PR has changed drastically and needs to be reconsidered.

Spacedong:
nice

Kokojos:
Not agreeing with the CLF3. 5 units of it sets you on fire for a long time, and the only thing that could save you is being right next to a pool

ghost:
It's easily countered by fire extinguishers.
Hell, I'd argue the good old bottle smash and light strat is more effective than CLF3.

And considering borgs used to have polyacid before it was nerfed, mediborgs were meant to be feared when emagged.
When I was testing it, it took a good while to crit someone with the spray too.

Kokojos:
Removed my previous post after testing. Nevermind me, this is gucci +1/3

ArcLumin:
Just remove nutriment drain on DD

Badmex:
Nah DD should stay that way. Also yeah seems nice that emagged Mediborgs are better now, still nobody will use them because emagged sec borgs for the win. +1/3

dreamcrab:
+33333333

KayeArray:
What about omnizine instead?

ghost:
> omnizine

Surely you are jesting
That's botany only dude.

Carbonhell:
Code ok, got approval, merging

Issue #2623, opened 2016-07-14
plasma tanks appear under tank storage units


Spacedong:
layer problems perhaps
https://vgy.me/Siok8V.png

Mithrandalf:
I've noticed this. It fixes itself sometimes when you take a step.

PR #2698, opened 2016-07-31
World's smallest layering bugfix

ghost:
Fixes https://github.com/HippieStationCode/HippieStation13/issues/2623

It's one line of code.

### DIFF
--- a/code/game/objects/structures/tank_dispenser.dm
+++ b/code/game/objects/structures/tank_dispenser.dm
@@ -5,6 +5,7 @@
 	icon_state = "dispenser"
 	density = 1
 	anchored = 1
+	layer = 2.9
 	var/oxygentanks = 10
 	var/plasmatanks = 10
 	var/list/oxytanks = list()	//sorry for the similar var names
--- a/html/changelogs/GandalfGreyhame - Tank Layering.yml
+++ b/html/changelogs/GandalfGreyhame - Tank Layering.yml
@@ -1,0 +1,36 @@
+################################
+# Example Changelog File
+#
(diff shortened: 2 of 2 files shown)


ghost:
This PR was an exercise in patience.

It's like trying to thread a needle with dicks for fingers.

Carbonhell:
I don't think this is the proper fix, you sure it is?

ghost:
Well I mean I tested it and it works.

What makes you think it isn't?

Pyko1:
Pretty much all the machinery is 3.0, I think it would be better to move the tanks above it.

ghost:
Pretty much everything in general is 3.0

I figure moving the dispenser down makes more sense than moving the tanks up, since dispensers dont tend to move much during the round at all.

carlarc:
yes please

Gandalf2k15:
Code looks good. I'll ask this one question, did you test it?

Gandalf2k15:
Holding off as per request.

Carbonhell:
Issue is not caused by the layer probably,because the bug appears to happen only with one tank dispenser,the one in atmospherics control room. Check if that dispenser's fixed with the layer edit.

ghost:
That's the one I tested.

I can tell you without a shadow of doubt that this will fix the issue, only question is whether you think this is the best way to go about it.

Carbonhell:
Mah, i don't get why this bug applies to a single tank dispenser. I still don't think this is a code issue,but rather a mapping issue,but if you claim it fixes it, then merging

PR #2699, opened 2016-07-31
Update towards the shema and config

Badmex:
Added HeadAdmin as a normal rank towards the shema. It was also added towards the basic config, that is being used without a db.


ghost:
Needs Nalar

Badmex:
Not even changing anything for us. Just for people that will set up a DB. Still Nalar will merge it and probably also add HeadAdmin as a real rank towards the server then.

Carbonhell:
@nalarmg

PR #2694, opened 2016-07-31
Makes radiation storms more blowout-esque

calibraptor:
Complete with screen shaking and fun sounds to scare the crap out of people experiencing it, also  the radiation is strong enough to significantly fuck you up if you don't get your ass into maintenance.

Only element I couldn't include was the entire station changing color as the storm blows through. Hoping someone more experienced can add that later on.

https://www.youtube.com/watch?v=UdXX2JDRhU0&feature=youtu.be

PR #2702, opened 2016-08-01
Revert "Makes radiation storms more blowout-esque"

Gandalf2k15:
Reverts HippieStationCode/HippieStation13#2694

Possible server fix. 
Only thing that I can see that would cause the server to die.


TDcoolguy300:
This is not an issue

PR #2703, opened 2016-08-01
Fixes a runtime with addictions

Kierany9:
For the record, I only have a vague idea of how what I did fixes the runtime, but it stops addictions from spewing out runtimes every tick and addictions now work properly. **Although I'd rather have this merged *after* we can isolate the cause of, and fix, the server stability problems we're having right now.** Also, I'd suggest reducing the duration of addictions as they are extremely crippling, have no cure and can take people out of the round for at least 10 minutes, or kill them if they can't get to the medbay on time. Not doing it here because I don't want a runtime fix to get boged with features that'll end up needing HA approval.

it just works


TDcoolguy300:
Bug fix, merging

PR #2705, opened 2016-08-03
Improve overlay code for taperecorder tapes

optimumtact:
This can cause pathological performance when the item is set on fire and fireact
is called repeatedly.

Fixes this by only adding the ruined tape overlay once.

See also https://github.com/tgstation/tgstation/pull/19667


TDcoolguy300:
A+ Optimum, never would of saw this myself. Seeing it as a fix i shall go ahead and merge it.

optimumtact:
Well like I said on my pr it was only through a combination of luck and a timely profile that it was found.

PR #2718, opened 2016-08-08
Reduces the price of the Wrestling Belt and CQC

Kierany9:
This was previously in the King of the Disk PR and didn't have much to do with the gamemode so I took it out.

Reduces the price of Wrestling and CQC from 12 TC to 5 and 7 TC respectively. I have no idea what Crystal was thinking when he made these 12 each, both are fun memey 1v1 items, neither of them can compete with the Esword or Power Fist in terms of usability or power.

### DIFF
--- a/code/datums/uplink_item.dm
+++ b/code/datums/uplink_item.dm
@@ -480,13 +480,13 @@ var/list/uplink_items = list()
 	name = "Tactical Gloves"
 	desc = "Comfortable grey gloves with the CQC martial art inside."
 	item = /obj/item/clothing/gloves/cqc
-	cost = 12
+	cost = 8
 
 /datum/uplink_item/dangerous/wrestling
 	name = "Wrestling Belt"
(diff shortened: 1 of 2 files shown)


Mithrandalf:
Kierany does it again.

10/10

GeneralGoose:
tru
also someone can be bothered enough to copypaste disguise code from the gloves onto belt and make it possible to disguise it as any other belt (fanny packs, medical and toolbelt), so it isn't an instant giveaway

Badmex:
Maybe cqc is still a bit too high? Overall yeah they are gimmik items that nobody is using because they cost so much tc. +1/3

calibraptor:
+1 I'd love to see them used more often.

GeneralGoose:
@Badmex CQC gloves can be disguised, belt can't, higher price makes sense

Nebyab:
ok

Kokojos:
+1/3 our savior is here, amen praise be to the lord Kienranyyyeyeneyen7

PR #2753, opened 2016-08-15
King of the Disk major bugfixes

Kierany9:
- Fixes a bug where where the disk would not grant extra telecrystals when stored in bags.
- Fixes a bug where mindslaves holding the disk would receive a message about gaining extra TC instead of whoever implanted them.
- The PDA uplink now automatically refreshes when gaining extra TC.

### DIFF
--- a/code/game/gamemodes/traitor/king_disk.dm
+++ b/code/game/gamemodes/traitor/king_disk.dm
@@ -37,24 +37,27 @@
 	return
 
 /datum/game_mode/traitor/king_disk/process()
-	for(var/obj/item/weapon/disk/nuclear/N in poi_list)
-		var/atom/disk_loc2 = N.loc
-		if(istype(disk_loc2, /mob))
-			var/mob/M = disk_loc2
-			if(M.mind)
-				var/datum/mind/traitor = M.mind
+	for(var/mob/M in living_mob_list)
+		if(M.mind)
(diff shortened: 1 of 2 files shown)


Earthykiller127:
You forgot one important problem, it's the gamemode itself

code/game/gamemodes/traitor/king_disk.dm
```diff
@@ -37,24 +37,27 @@
...
 						N.king_timer = 0
-						var/list/all_items = traitor.current.GetAllContents()
+						var/mob/H = M
 						if(traitor.special_role == "Mindslave")
 							for(var/datum/objective/protect/P in traitor.objectives)
 								if(P.target && P.target.current)
 									all_items = P.target.current.GetAllContents()
+									H = P.target.current
 						for(var/obj/item/device/uplink/U in all_items)
 							U.uses += 1
```
JohnGinnane: Does this mean if I have multiple traitor PDAs they'll all get one extra use?

Kierany9: Yep

JohnGinnane: Neat

JohnGinnane:
Code looks good, merging as they're bugfixes

PR #2776, opened 2016-08-22
Brig fixes

calibraptor:
- Removed mimic from maintenance
- Replaced fake eguns in armory with real ones
- Fixed some wires


Kokojos:
Yes yes yes merge this already,

GenericGoose:
kek all those fails

PR #2823, opened 2016-08-29
Small change to make strangling a bit more viable. Death now results in

ghost:
about 20 ticks, rather than about 100 before.

PR #2827, opened 2016-08-29
Combatchanges

ghost:
As before, a change to make strangulation more viable. Does not modify CQC or garroting, as I feel they were already in a good place. My original change brought the DME file from my local directory in, this PR will not, although being used to subversion did cause a few headaches for me :P

Original PR was #2823


GenericGoose:
cool

ghost:
The overwhelming consensus seems to be that a buff to strangulation is a good thing for balance. Hopefully someone who can merge it in will agree.

GenericGoose:
since getting an agressive grab will result in a kill, because target can't break it, can make strangling faster as well

SplergeBucket:
yeah we can see how it goes

Kierany9:
I'm not so sure on this. While the old choking time is way too long, 20 ticks until death(which means ten ticks until crit) is _really_ short.

ghost:
@kierany9 please recall that you are standing is one place for 40 seconds and your victim can still scream so this is a much needed buff

Kierany9:
You can always just muffle them by aiming at the mouth, which prevents screaming and talking as long as you have a neck grab. I know the old time is shit but there's a really interesting concept called a happy medium that would probably be good here.

Kokojos:
Making it 5x faster seems like too much to me. How much time do you need to strangle someone from 100% to death?

ghost:
i would like the admins to record how many deaths occur from stranglation each round and consider whether or not this is too low

Kierany9:
There's a difference between getting something just right and hilariously overbuffing something. You honestly should know the difference.

ghost:
I felt that considering the time it takes to actually get into a neck hold, this was a very fair happy medium. As it stands, this is hardly viable unless you can get someone still for a prolonged period of time anyway. The difficulty is all in establishing the grab. This just makes the process actually faster.

I think it should be playtested, but I'm willing to bet hard that this strikes a fair balance. If problems arise in-game from it, then it can obviously be tweaked, but I'm willing to put 20 bucks on this being perfect ;)

ghost:
@Kokojos since the majority of the difficulty is in establishing the neck hold to begin with, I feel this is about right. If the overwhelming majority believe I should, I'll reduce it from a 5 per tick to a 3 per tick, which will result in approximately 33 seconds rather than 20. I just feel that's a bit too long considering the difficulty of establishing the grab.

ghost:
Honestly I was never around for the quickchoke meta.
But I get the feeling that it was nerfed to the ground for good reason.

ghost:
I think with the buffs to CQC this could go into pretty strong territory but I guess I'm willing to trial it?
It needs a changelog though. +1

Kokojos:
At the worst, we can see how it goes :+1: 22 seconds seems okay with me.

ghost:
Added a changelog as @McBawbaggings requested.

TDcoolguy300:
3 HEadmin approval, code is fine, blah blah merging.

PR #2847, opened 2016-08-31
Fixed the bug referenced in #2846 regarding nukeops.

### LINKED ISSUE #2846: You can declare and land at the same time
Title. If you're fast you can take the drop pod, have it charging up to go, and right before it leaves, declare. The pod will still drop and you'll have your 200 TC

ghost:
Fixes #2846


ghost:
looks good 10/10

Issue #2841, opened 2016-08-30
Bomb wrapped in package explodes as normal, keeps package intact


ghost:
Reproduction:
- Grab a bomb
- Prime it
- Use packaging wrapper on it
- Watch it explode
- Unwrap the package and witness the disappearing bomb

ArcLumin:
This is true for many items, it doesn't qdel the outer shell/container. Package wrapper isn't the only issue, it's just one that isn't commonly blown up by the bomb

PR #2849, opened 2016-08-31
Addresses Issue #2841

ghost:
Addresses Issue #2841

### DIFF
--- a/code/game/objects/explosion.dm
+++ b/code/game/objects/explosion.dm
@@ -14,9 +14,11 @@
 
 
 /proc/explosion(turf/epicenter, devastation_range, heavy_impact_range, light_impact_range, flash_range, adminlog = 1, ignorecap = 0, flame_range = 0 ,silent = 0)
+	var/obj/item/S = src
+
 	var/obj/item/weapon/storage/explosion_container = null
-	if(istype(src.loc, /obj/item/weapon/storage))
-		explosion_container = src.loc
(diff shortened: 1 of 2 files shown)


ArcLumin:
I wonder how exactly this would work with very small bombs? If it's deleting the container rather than doing the explosion (emptying before exploding), wouldn't that cause small bombs (i.e. PDA explosions) cause say, the backpack they're in to be entirely deleted, or even worse, delete the mob that's holding them?

ghost:
With smaller bombs, if they're big enough to harm anything, they're big enough to destroy the container they're in, realistically speaking. But, as noted, I've made it drop all items from the container before destroying it. This has the added effect of stuff inside the container being able to be pushed around, a la shrapnel when the explosion goes off.

And it checks if the container is of type /obj/weapon/storage. It would never destroy the player.

ArcLumin:
Yeah, saw that /obj/weapon/storage after I posted that.

I dunno off the top of my head, is package wrapping part of /obj/weapon/storage?

ghost:
The boxes it produces are, yeah.

ArcLumin:
They're not /obj/item/smallDelivery and /obj/structure/bigDelivery?

ghost:
/obj/item/weapon/storage/box

ghost:
You were right about the delivery objects being the issue. Changed to address those instead.

TDcoolguy300:
Code is good, fixes a known issue. Merging.

Issue #2848, opened 2016-08-31
Wizard rounds don't end properly.


Spacedong:
Once the wizard dies it never mulligans nor ends the round, so its just stuck in a period of extended mode unless wizard had summoned guns causing there to be survivors.

Mithrandalf:
It spawns fresh antags doesnt it

Spacedong:
I just said, mulligan never happens. No, it doesn't.

McBawbaggings:
Mulligan was intentionally disabled IIRC.

Alphora:
Well, I'm willing to fix this, but what's the desired behavior? Immediate round end? Shuttle call? New antags?

Kokojos:
It should trigger Mulligan IMO.

Alphora:
No idea what mulligan is.

Mithrandalf:
she's that chick from that one fighting game

Kokojos:
It automaticly makes new antags if more than 70% of the crew isin't dead. It's activated after a malf AI loses if in the remove blob gamemode if they blob dies early IIRC.

Alphora:
So there's an already-existing proc? Or you're saying that's the behavior we want?

Kokojos:
It's already existing but I think it's more than just a proc. I honestly have no idea where it's stored.

McBawbaggings:
it's a config setting IIRC
I think people are saying wiz rounds should end on wizard death?

Alphora:
I'll find it or make a new one.

ArcLumin:
Don't get it confused with the mulligan chemical

ArcLumin:
Not an issue, please close

PR #2850, opened 2016-08-31
Mulligan4wizards

ghost:
Addresses Issue #2848

### DIFF
--- a/code/game/gamemodes/wizard/wizard.dm
+++ b/code/game/gamemodes/wizard/wizard.dm
@@ -178,6 +178,8 @@
 		if(isliving(wizard.current) && wizard.current.stat!=DEAD)
 			return ..()
 
+	round_converted = convert_roundtype()
+
 	if(SSevent.wizardmode) //If summon events was active, turn it off
 		SSevent.toggleWizardmode()
 		SSevent.resetFrequency()
--- a/html/Zach - wizardmulligan.yml
+++ b/html/Zach - wizardmulligan.yml
@@ -1,0 +1,4 @@
+author: Zach
+delete-after: True
+changes:
+  - tweak: "The server now mulligans when the wizard dies, rather than just being a boring pseudo-extended round."


TDcoolguy300:
This needs input from the Headmins, as this was intentionally disabled originally.

Kokojos:
I am okay with this as wizard often die early. :+1:

danhibiki:
There is a _huge_ disdain for the mulligan system. I don't understand it, as it just keeps a round going for as long as the station can support it really.

ghost:
It was decided at a meeting to disable mulligan on wizard.
Honestly I think the classic round is over when wizard is dead is the best way to do it.

ghost:
Only problem with that bawb, is that staff of transformation 'kills' wizards if they lose their hands.

With mulligan, the round doesnt end and they can usually get themselves transformed back.

ghost:
Also
https://github.com/HippieStationCode/HippieStation13/blob/0a9adaeb95bb09a8e50bec0df885bebd40f2ed80/config/initial_config/game_options.txt#L102

https://github.com/HippieStationCode/HippieStation13/blob/0a9adaeb95bb09a8e50bec0df885bebd40f2ed80/config/initial_config/game_options.txt#L127

It's a config option, new code shouldn't be added.

ghost:
I should say I would like the round to end on wizard death.

SplergeBucket:
i dont get the issue with mulligan
just make mulligan for wizard rounds
it's the easiest solution

SplergeBucket:
having the whole thing end on wizard death could be cool but when you've got the issue of survivors or apprentices it can all get a bit clusterfucky and in that case i dont see why mulligan shouldnt be a thing

danhibiki:
Its actually quite easy to do that within the current configuration, I think its just turned off. We'd need someone with box access to confirm

Nebyab:
mulligan is bad and you should feel bad

also mulligan is a config option which needs to be changed on the box

just make apprentices and survivors not count for round ending imo. Wizman is the only one who matters.

SplergeBucket:
> mulligan is bad
> t. doesn't play the game

TDcoolguy300:
Due to this being a config issue, closing.

PR #2861, opened 2016-09-03
Fixes bolt of change turning people into drones without restrictions.

ghost:
They could fire guns and attack people and shit.
Also adds a comment to warn people not to use /mob/living/simple_animal/drone

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/drone/_drone.dm
+++ b/code/modules/mob/living/simple_animal/friendly/drone/_drone.dm
@@ -10,6 +10,7 @@
 #define REPAIRDRONE	"drone_repair"
 #define SCOUTDRONE	"drone_scout"
 
+//Please don't use this, use /mob/living/simple_animal/drone/safedrone as they have restrictions.
 /mob/living/simple_animal/drone
 	name = "Drone"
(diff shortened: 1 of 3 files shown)


danhibiki:
This is a relatively urgent fix

Spacedong:
but the real question is why does /mob/living/simple_animal/drone even exist at this point

blobbernaut:
:-1: I think it should be a lawless drone from the staff of change to be honest, but maybe it should have a different sprite

myontasm:
Blob, we already have the chance to score a syndieborg from the staff of change. That's enough in my opinion.

ghost:
The reason /mob/living/simple_animal/drone exists is so I wouldn't have to override onclick code for syndrones and safedrones.

ghost:
Staff of change should be filled with lots of bad choices and few good choices. It should be almost always bad. Lawless drones is bad.

SplergeBucket:
this is a fix
should be merged

ArcLumin:
I think it's fine as is

Bigger issue is staff of changing drones into other mobs

myontasm:
Why isn't this merged yet aaaa

TDcoolguy300:
¯_(ツ)_/¯

PR #2873, opened 2016-09-05
Fixes #2866, talking while in crit will now parse HTML properly. Also fixes chatrooms having the same problem.

### LINKED ISSUE #2866: When text is italicized, an apostrophe turns into &#39;
This ruins the moment so much! It's basic text handling stuff, but it really gets annoying when you are in crit and trying to talk to people.

TrashFiend:
Fixes #2866

Before, any time someone talked while in crit or whispered their last breath, an apostrophe would show up as &#39; . This fixes it from my testing, nothing else happens that's unintended. This applies for both talking in crit along with last words stuff.

Also, the problem was the same in the chatroom, so I fixed that too. It's hack-y as shit but nobody uses the chatroom anyway and it is totally a fix.

### DIFF
--- a/code/game/objects/items/devices/PDA/chatroom.dm
+++ b/code/game/objects/items/devices/PDA/chatroom.dm
@@ -49,7 +49,7 @@ var/list/chatchannels = list(default_ntrc_chatroom.name = default_ntrc_chatroom)
 			else
 				return "BAD_COMMAND"
 
-	return send_message(client,nick,message)
+	return send_message(html_decode(client),html_decode(nick),html_decode(message))
 
 //the following are helper procs, FOR INTERNAL USE ONLY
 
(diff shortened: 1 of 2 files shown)


ghost:
Can't you code inject with this?

TrashFiend:
Not that I've seen. It follows the same coding as announcements and a few other mobs, and there hasn't been problems with those. I'll do a more thorough check in a second.

ghost:
I mean if this is safe then good shit.

TrashFiend:
If you have a proper example to test with that would be solid, but it seems to work.

ArcLumin:
RIP & #39;

TDcoolguy300:
From what i remember HTML decode should sanitize the text, and if not there are other checks elsewhere.

PR #2880, opened 2016-09-06
Toggle PDA flashlight verb

Theoman02:
Made it so that you can now toggle the PDA's flashlight either from the menu on the right or by right clicking the PDA itself.


ArcLumin:
Why not try middle clicking to turn on the flashlight? MiddleClickOn will let you simply middle click your PDA to turn on the flashlight

Spacedong:
ehh I like opening the PDA and switching it on, just gives the sense that you're actually using it

Theoman02:
Tried to ArcLumin, but it seems to be only for mobs, and it also makes you switch your active hand as well. Would need some rewriting in click.dm, so I'm just sticking with the simple option for now.

myontasm:
Thank you, this is something that I've wanted for a while.

SplergeBucket:
but i right click the PDA itself to remove my ID

doobiedoo23:
I like QoL changes as much as anyone else, but I think having to open your PDA to toggle it is better.

-1/3

Kokojos:
Useless to me.

ghost:
I like it. +1

Nebyab:
opening your PDA is IMMERSIVE

also its just a generally more appropriate way to handle it, not everything needs to be verbiaized

SplergeBucket:
2 Headmins, and 2 GMs against it.
Safe to deny this really.

Pyko1:
apparently denied, closing

PR #2899, opened 2016-09-11
In case of emergency merge PR

Kierany9:
I'll be gone for a couple of weeks and won't be able to code during that time. Since the magboot stomps are somewhat experimental in balance, there's a chance they may be horrendously OP. If they are, please merge this PR while I'm away.

- Reduced damage from 45 to 40.
- Increased windup time from 4 seconds to 4.5 seconds.
- Increased cooldown on hit from 0.5 seconds to 1.5 seconds
- Increased cooldown on miss from 0.8 seconds to 1.5 seconds


ArcLumin:
I don't think this is needed

IvanTheCracka:
wtf its almost neglibigle values reee

Kokojos:
No need for this.

PR #2909, opened 2016-09-13
Allows the chaplains' null rod to be reskinned and renamed

ghost:
Adds the framework for weapons to be renamed and reskinned.

axe
fish
sword
sbf
sbb
sbf2
srh
ab
ab2
arh
fb

fb2
frh


ghost:
Oh, and I'm dumb as a sack of bricks so I couldn't figure out how to add the sprites the easy way.

SplergeBucket:
wot do the skins look rike

ArcLumin:
You made your "wopit" give names for the "wopit" when reskinning. Make it use [name] rather than the word wopit

ghost:
References to "wopit"s have been removed, and replaced with [name].

A couple of belt sprites were also backwards. Had to actually look pretty close to notice it, but they've been fixed.

ArcLumin:
Do you have an image of the reskins? Put them in the opening paragraph

ghost:
Images of reskins added

Hackole:
The sprites look good and the idea is neat, +1/3

ghost:
One thing I did was change the weapon weight from 1 to 3, because sticking a sword, pole axe, or great big fish in your pocket seemed a little odd. I'm wondering if it would be acceptable to give a minor damage buff as well, to compensate.

danhibiki:
I don't think the null rod remotely needs a buff, but wouldn't setting it to weight to also stop you from pocketing it, but allow you to put it in backpacks?

ghost:
Weight class three allows the item to fit into bags, but not boxes. I can change it to weight class two if you want it to fit into boxes though.

ArcLumin:
I'd love to see the different skins get different properties

ghost:
I tried, but it didn't work. I really wanted to give the bladed skins their own hitsounds. The method I used was just creating a new child item for the null rod for each of the options.

If anyone has another idea that might work with the way this is set up, I can give it a go.

danhibiki:
If each one is a child, it would be as simple as assigning the new values to the children, right? In general, I don't think "reskins" should entail buffs though, as the idea of a reskin is to keep the original functionality, but change the look. Its pretty much semantics i'm talking about though

ghost:
Yeah, I wanted to keep this visual only, and sound if possible. Really, my only concern at this point is getting the PR processed, and I don't much care what's done or not done to it.

danhibiki:
Sound shouldn't be hard, that should be just defining hitsound for each child.

code/game/objects/weapons.dm
```diff
@@ -2,6 +2,32 @@
...
+
+/obj/item/weapon/proc/reskin_wopit(mob/M)
+	var/choice = input(M,"Warning, you can only reskin your [name] once!","Reskin [name]") in moptions
+
+	if(src && choice && !M.stat && in_range(M,src) && !M.restrained() && M.canmove)
+		if(moptions[choice] == null)
+			return
+		else
+			icon_state = moptions[choice]
+			item_state = moptions[choice]
```
danhibiki, line 26: Around here is where you'd want to reset your weapon sounds. I'd just organize the soundbits to be named after their moptions name, and do something like "hitsound = '/sound/weapons/[moptions[choice]].ogg'" Not sure if you can use that syntax specifically for that function, but I'm fairly certain its possible. Or doing a length if statement, but that isn't wanted I assume.

ghost:
I've got something I'll try out

ghost:
Well, I can set up an if statement in weapons.dm if the hitsound thing is important, but it'll be ugly.

Unless something can be found to make Dan's idea work, and we're both stumped at the moment, I'd rather let it be.

ghost:
Changed the left inhands to match the right ones. I always forget to do that.

I've also given up completely on giving the skins unique hitsounds. It would require making a very large mess, which is the exact opposite of what I'm trying to do here.

LuisReyesX:
I would like the possessed sword from TG to be ported.

doobiedoo23:
I'm pretty sure /tg/ did this already, might be worth checking out their code to see if their solution was similar to yours?

Anyway, +1/3

ghost:
/tg/ has the code to rename weapons in weaponry.dm, versus weapons.dm where I have it.

Nothing under weapons for reskinning though.

Kokojos:
Pretty nice +1

Kokojos:
Testes live yesterday

PR #2910, opened 2016-09-13
Allows Surplus Crate to be purchasable during Nuke Ops or KotD

ArcLumin:
It'd be a fun gimmick nuke ops to just get bulk random items, and you'd need a lot of time with the nuke disk to get the surplus crate in KotD

### DIFF
--- a/code/datums/uplink_item.dm
+++ b/code/datums/uplink_item.dm
@@ -1248,7 +1248,6 @@ var/list/uplink_items = list()
 	desc = "A crate containing 50 telecrystals worth of random syndicate leftovers."
 	cost = 20
 	item = /obj/item/weapon/storage/box/syndicate
-	excludefrom = list(/datum/game_mode/nuclear,/datum/game_mode/traitor/king_disk)
 
 /datum/uplink_item/badass/surplus_crate/spawn_item(turf/loc, obj/item/device/uplink/U)
 	var/obj/structure/closet/crate/C = new(loc)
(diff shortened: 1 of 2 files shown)


ArcLumin:
Also, this lets some nuke ops teams use items that are underused/would never be used otherwise

Hackole:
Don't like this, nukeops being able to buy surplus crates would be a little bit too overpowered in my opinion since you could just buy an assload of them, and that would give them a lot more tc than they have worth of equipment.

ArcLumin:
Well, 90% of what you get in a surplus crate is crap or stuff you'll rarely use, so I dunno if using only 10% of what you get is gonna be super viable for em

doobiedoo23:
+1/3

ghost:
KotD being able to buy that is not good imo.
Nuke ops less so but still a bit iffy.

ArcLumin:
Well, you need to get 10 more TCs than what you start with to buy it

Apparently emagging cargo was removed in KotD and replaced with 14 pinpointers, but you still need an extra 10 tc's to buy a surplus

Kokojos:
Hehh I don't see the point in this.

Nebyab:
Giving nukies surplus crates seems like a bad idea.

SplergeBucket:
yeah, shit ton of nuke spam shit
announce assault
buy 300 crates

SplergeBucket:
no i say
no

Lawligagger:
As funny as it is no

PR #2923, opened 2016-09-21
Adds Bananium, Fugus to mining

ArcLumin:
Adds Wumborian Fugus and bananium to mining. Also veeeerrry slightly boosts chance of mining mobs spawning, to compensate for the new mob.

### DIFF
--- a/code/modules/mining/mine_turfs.dm
+++ b/code/modules/mining/mine_turfs.dm
@@ -91,6 +91,7 @@ var/global/list/rockTurfEdgeCache
 		"Uranium" = 5, "Diamond" = 1, "Gold" = 10,
 		"Silver" = 12, "Plasma" = 20, "Iron" = 40,
 		"Gibtonite" = 4, "Cave" = 2, "BScrystal" = 1,
+		"Bananium" = 1,
 		/*, "Adamantine" =5*/)
 		//Currently, Adamantine won't spawn as it has no uses. -Durandan
 	var/mineralChance = 13
@@ -320,7 +321,7 @@ var/global/list/rockTurfEdgeCache
 
 /turf/simulated/floor/plating/asteroid/airless/cave
 	var/length = 100
-	var/mob_spawn_list = list("Goldgrub" = 1, "Goliath" = 5, "Basilisk" = 4, "Hivelord" = 3)
+	var/mob_spawn_list = list("Goldgrub" = 1, "Goliath" = 5, "Basilisk" = 4, "Hivelord" = 3, "Fugu" = 2)
 	var/sanity = 1
 
 /turf/simulated/floor/plating/asteroid/airless/cave/New(loc, var/length, var/go_backwards = 1, var/exclude_dir = -1)
@@ -399,7 +400,7 @@ var/global/list/rockTurfEdgeCache
 	if(prob(30))
 		if(istype(loc, /area/mine/explored))
 			return
-		for(var/atom/A in ultra_range(15,T))//Lowers chance of mob clumps
+		for(var/atom/A in ultra_range(12,T))//Lowers chance of mob clumps
 			if(istype(A, /mob/living/simple_animal/hostile/asteroid))
 				return
 		var/randumb = pickweight(mob_spawn_list)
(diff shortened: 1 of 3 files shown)


ArcLumin:
Done a few surveys in OOC and on the slack, player interest into adding these to mining is overwhelmingly positive, so I figured I'd pop em in

ArcLumin:
For those of whom are unaware what Fugus are, here's a breakdown

Small, weak mob normally, puffs up to be a very large mob when threatened.
When large, they can deal tons of damage, but they can only be large for a short time, and there's a cooldown afterwards
They'll try to escape while small.
They are fully coded, with multiple sprites, and verbs for if they become player controlled.
They drop an item that allows you to buff simple mobs (provided they already haven't been buffed). This makes their sprite larger, increases max and current health by 50%, adds up to 10 extra damage to the min and max, and allows them to break environmental objects.

ghost:
I think fugu can be used on drones, cultie mobs and holoparasites.
There's a list in the fugu code banned_mobs that is empty.
Provided they get banned from use on cultie, drones and holoparasites and any other mobs along those lines then sure why not.

ArcLumin:
I don't see any issue with drones. They can't directly attack (and their laws don't allow them to) so that's fine. The extra health is negligible since they have shit HP anyways. They can't break walls and shit since they have no harm intent. Effectively, it only makes them a bigger target.

Cult stuff is also not an issue, since they die pretty fast anyways, and it'd be far too costly to be popping them on every harvester or jugg around. 

I'm 99% sure that the buff actually gets removed when the holopara returns to its owner, and holoparas themselves have no health so it's fine. I'll have to double check if that's the case for sure though.

I'll double check the holopara one, and so long as it gets removed when the owner returns, I don't see an issue with them being affected by the fugu stuff.

ghost:
It gives drones more HP, which I dislike.
Cult stuff is an issue very much so. You're doubling jugger damage from 30 to 60. HP from 250 to 325.
Double damage holoparasites is also retarded and shouldn't occur.
resummoning holos doesn't make new holos, it just changes their location so I don't see why it should "reset".

ArcLumin:
It's not double damage though. The damage buff is capped at +10

ghost:
> Max is 10
>     A.melee_damage_lower = max((A.melee_damage_lower \* 2), 10)
>     A.melee_damage_upper = max((A.melee_damage_upper \* 2), 10)
> Max

ArcLumin:
Isn't the 10 there the max amount it'll add? It'll do upper and lower \* 2 but go no further than 10, right?

ghost:
Okay
If a function is max(30, 10) you will get the MAXIMUM VALUE within the arguments, that being 30.

ghost:
To add, the 10 is there in the case harm intent did like 3 damage, it wouldn't be a measly 6 but at least 10.

ArcLumin:
Ah, I see. I was mistaken. It was the other way around.

I can see the point for blacklisting the cult mobs, then.

ghost:
I'd wait for other headmins to chime in before making changes though since I might be in the minority.

ArcLumin:
Added in blacklisting of Holoparas, Constructs, Drones and True Changelings

doobiedoo23:
Yeah booooooooooooooy

ArcLumin:
WHERE MY OKEYS

@EVERYONE

Nebyab:
I appreciate the vague spongebob reference.

This seems neat.

ArcLumin:
Two head okeys? So it's ready to merge?

SplergeBucket:
ok

ArcLumin:
Let's MEEEERRRRGE

FUSIOOON HAAAAA

blobbernaut:
So: I don't think drones should be on the blacklist. 

For holoparas, keep in mind that the health boost does nothing for them, it'll only make them larger and give them more damage. 

For constructs, keep in mind a juggernaut will be so large it'll be really hard to even touch anything with a melee attack because you'll just click on yourself, so it's probably bad on them. It'd be nice for an artificer to have, but it's not game breaking. It should be rewarded. 

For true changelings, the same problem as the juggernaut. They'll do more damage sure, but they'll also be easier to hit and have a lot harder time hitting anything else due to their own sprite blocking whatever they're trying to click.

ghost:
Drones, holos and culties don't even make any thematic sense though

blobbernaut:
You can use bruise packs on a drone or a construct too. Also wizards exist. 

Speaking of which, fugu implant into your body give you bonus to max health and punch damage?

ArcLumin:
I could implement that, although it'd likely be in a later PR

Let's get this merged first, it has all the okays and the code's good. WHERE ARE YOU MERGERS

ghost:
Wizards have suspension of disbelief.
Bruise pack healing simple mobs is very old legacy code.

Just because current code has issues doesn't make it an excuse to dismiss making it worse. ;)

Pyko1:
code looks fine and has enough headmin votes

PR #2939, opened 2016-09-26
The Makarov Update

ghost:
Adds the Makarov PM and PB.

The Standard Makarov hold 8 rounds of 9mm. Standard magazines are loaded
with AP, along with toxin and incindiary as special ammo.

The Makarov PB is the same as the standard makarov, but with an integral
suppressor. It can also be named with a pen.

Added a syndicate bundle containing the standard Makarov, three standard
magazines, a combat knife, soviet uniform, combat boots, conbat gloves,
ushanka, agent ID, voice changer mask, and a bottle of vodka. Minus the
clothes, the pack is worth 19 TC. With the clothes, probably closer to
25.

Added a mime only uplink kit containing a Makarov PM, two toxin mags,
and one incindiary mag for 14 TC. 10 for the gun, 2 for the toxin mags,
2 for the incindiary mag.


ArcLumin:
Bundles should really be discounted overall since they're, you know, bundles.

Also maybe make it an armored version of those clothes?

code/modules/projectiles/ammunition/magazines.dm
```diff
@@ -448,6 +448,22 @@ obj/item/ammo_box/magazine/tommygunm45
...
+	icon_state = "makarov"
+	max_ammo = 8
+	multiple_sprites = 2
+
+/obj/item/ammo_box/magazine/makarov/mktx
+	name = "Makarov magazine (Toxin-Tipped 9mm)"
+	ammo_type = /obj/item/ammo_casing/c9mmtox
+
+/obj/item/ammo_box/magazine/makarov/mkinc
+	name = "Makarov magazine (Incindiary 9mm)"
```
Kierany9: Incendiary

Kierany9:
>It's another guns with snowflake ammo episode.

Why is this gun mime-only and why should I use this over the Sketchin if it costs twice as much, does less damage and has the same ammunition capacity?

>Minus the clothes, the pack is worth 19 TC.

The entire point of bundles is to provide more than 20 TC of equipment but lock you into a certain playstyle due to the specific combinations of gear that you get. This bundle is just a worse version of the Snake Bundle. If you want to add the Makarov to a bundle, perhaps make it a 50/50 chance to spawn instead of a sketchin or glock in one of the other bundles.

PR #2941, opened 2016-09-26
Makes Life Reagent spawn by volume

ArcLumin:
Rather than spawning 1 regardless of amount, it now spawns a mob dependant on how much is made.

It'll prolly fail but hey


ArcLumin:
Closing this, I have a better idea I'll try later

PR #2929, opened 2016-09-24
fixes giant tajaran species typo

Spacedong:
its spelled "tajaran" not "tarajan" you fuckin idiot whoever coded this

PR #2953, opened 2016-09-30
Revert "fixes giant tajaran species typo"

tortellinitony:
Reverts HippieStationCode/HippieStation13#2929

Cats are invisible now ffs


TDcoolguy300:
> Spacedong doesn't realize that icons are dependent on species names

Carbonhell:
Why do we even let spacedong code

PR #2955, opened 2016-10-01
fixes invisible cats

Spacedong:
merge this immediately after the revert of the revert then update after both are merged


Pyko1:
is this tested?
I assume no, since its in 2 parts.

Pyko1:
No missings tails or ears?

Seeing the onyl change is to the body, and noting to mutated bodyparts.

QuarianCommando:
Doesn't look like there's any issue. Catbeast ears and tails aren't separate sprites like with moth wings. The revert being separate from this fix is proper protocol. 

The original revert should never have happened in the first place. A maintainer should have noticed the sprites weren't updated, and should have held off merging until this was added.

Pyko1:
dongdong catlove Part 2 take 2

Pyko1:
Aaaaand still broken...

PR #2998, opened 2016-10-08
Alien Fun Time

ghost:
Gives the AI controlled alien drones a verb, allowing players to morph
into humanoid drones should they be transformed by a wand/staff of
change, or stuffed into an AI controlled drone by a xenobiologist.

Allows alien infestations to happen more often, seeing as they almost
never happen, or happen so late the shuttle has arrived before the
queen's first wave of larvae can hatch.

Everything has been run through a test server, and it works just fine.

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/alien.dm
+++ b/code/modules/mob/living/simple_animal/hostile/alien.dm
@@ -42,6 +42,15 @@
 	var/plant_cooldown = 30
 	var/plants_off = 0
 
+/mob/living/simple_animal/hostile/alien/drone/verb/Awaken()
+	set name = "Awaken"
+	set category = "Alien"
+	set desc = "Awaken and gain full control of your body."
+	var/mob/living/carbon/alien/humanoid/drone/new_xeno = new (usr.loc)
+	usr.mind.transfer_to(new_xeno)
+	qdel(usr)
+	return
+
 /mob/living/simple_animal/hostile/alien/drone/handle_automated_action()
 	if(!..()) //AIStatus is off
 		return
--- a/html/changelogs/QuarianCommando - AI Alien Verb.yml
+++ b/html/changelogs/QuarianCommando - AI Alien Verb.yml
@@ -1,0 +1,6 @@
+author: QuarianCommando
+ 
+delete-after: True
+
+changes: 
+  - rscadd: "Added a verb to AI controlled alien drones, allowing players to morph into the humanoid drone type and granting access to all normal drone abilities."


ghost:
> Every xenobio traitor round is now a xeno infestation round.

I mean, there is good reason why xenos are a rare occurrence here.

ghost:
~~Also no changelog ;)~~

ghost:
wow thanks github for not showing that when I viewed it despite it being pushed.

ghost:
But xeno rounds are fun. And I added the change log already.

ghost:
The verb is only given to the drone as well, so no other xenobio or wizard aliens get to run around with their powers.

It would also take a xenobiologist around the same amount of effort to get a drone as it takes a decent miner to get himself hugged. Only time the drones would cause real chaos is during wizard rounds.

Spacedong:
>seeing as they almost never happen

gee

#i wonder why

ArcLumin:
I'm down for this

GuyonBroadway:
I think aliens need a balance tweak before they get made more common, as it stands it takes a full 17 mins from a hugger hitting a face to a fully grown alien becoming avaliable, larva having to sit around for 7 mins doing fuckall is really really dull. 

But then there's the fact that aliums can chainstun two guys with their super duper knockdown disarm thingy, this needs to be replaced and alien growth needs to be sped up. 

Aliens DO need a mechanism to get a human down long enough for their helmet to be removed and their face violated, I reccomend something with a longish CD but is a hefty one man stun, some kind of tail swipe that hits one target and puts them on their backside for a while. If you stick with a buddy you should be able to survive a lone alien, the aliens should have to group up and swarm to be effective.

Gandalf2k15:
Code check passed. 0/4 headmin input. PLS MAN GOD.

LuisReyesX:
Just remove the INFINITE MELEE STUN the aliens have and it might be fine

ghost:
+1 if xenos get nerfed

ghost:
So should I bundle a CD onto alien abilities that don't have them?

GuyonBroadway:
The issue is their "disarm" is a guaranteed stun and works much like the telebaton with a faster attack time. 

Sure, the ayys NEED something to keep a guy down long enough to remove his helmet without killing them in order to get their faces molested to continue the life cycle. But as it stands they could keep like... three guys knocked down getting a claw hit on each one before stunning them again.

A shower thought I had on the matter would be to give them the ability to "bite" someone and inject them with a mix of chems that keep them alive but unconscious. That way Xenos could wail on a target till they get crit, inject them with the chems and drag em on home without fear of accidentally killing them. Now that the Xeno's do not have a free GG-no-Re stun Humans have a chance to fight back. Perhaps consider giving their claws some armour piercing so they can reliably take an unrobust human down and call it a day.

ghost:
I variation of your shower thought could be allowing them to inject salbutamol to keep their victims alive through crit, or even to make a snowflakey alien chem that works wort of like formaldehyde, but with salbutamol or atropine.

The problem with aliens relying on crit is that this is Hippie, and people like to murderbone.

What could be done is to change disarm to deal stamina damage, maybe a little more than the boxing gloves. It would also give captured crew a chance to escape if left alone too long instead of being in crit. A chem that spawns atropine in their system would also bump them out of crit after a while.

One or both of those, I guess.

doobiedoo23:
-1

blobbernaut:
Facehuggers should inject epinephrine and salbutamol

SplergeBucket:
no

PR #3008, opened 2016-10-14
Removes second stun resist from antistuns

Kierany9:
Basically, after I reworked antistun chemicals, they allowed you to get up from a stun immediately after X ticks of not being stunned. However, after initial testing by our local powergaming expert showed that this was somewhat underpowered, I added a second stun resist that would allow you to resist stuns immediately if you spent 3 ticks unstunned after the timer hit X, _and said stun resist would only set the timer back to X, meaning you could ignore taser-strength stuns every three seconds regardless of the effectiveness of the chemical_. I think you can all see what went wrong here and why I'm removing this feature.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents.dm
+++ b/code/modules/reagents/Chemistry-Reagents.dm
@@ -133,10 +133,7 @@
 		metabolization_rate = 2 * initial(metabolization_rate)
 		for(var/datum/reagent/R in M.reagents.reagent_list)
 			if(R.stun_timer >= R.stun_threshold)
-				if(R.stun_timer >= R.stun_threshold + 3)
-					R.stun_timer = R.stun_threshold - 1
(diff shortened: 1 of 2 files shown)


trifornt-tinfornt:
undo adrenals nerf
otherwise yis

Hackole:
+1/3, this is absolute hell

McBawbaggings:
Yeah okay

Kierany9:
The effect this will have on adrenals is almost negligible considering how you use them. If anything it's a big buff by proxy by toning down the ridiculousness on the alternatives.

ArcLumin:
sounds good to me

doobiedoo23:
+1

ArcLumin:
Code's good.

SplergeBucket:
+1

PR #3018, opened 2016-10-26
Fixes various cooking recipes

ghost:
Fixes meatloaf being human meat only, makes a bunch of recipes require meatslabs rather than generic meat.


ArcLumin:
Code's good. Merging as it's a bugfix

PR #3021, opened 2016-10-30
Fixes gamma derelict having normal drones and not safedrones.

ghost:
Thus gamma derelict drones had no restrictions.


ArcLumin:
BUT GAMMA DERELICT IS THE LAND OF NO RUUUUUULES! 

code looks fine, merging after travis because bugfix

PR #3034, opened 2016-11-11
~ Balances the round start stim virus which its anti stun and huge sp…

Swagile:
~ Balances round start anti-stun and massive speed boost to a hardsuit / heavy equipment / damage no slowdown nerf.

~ Adds a changelog for users to know what has happened.

### DIFF
--- a/code/datums/diseases/advance/symptoms/stimulant.dm
+++ b/code/datums/diseases/advance/symptoms/stimulant.dm
@@ -10,7 +10,7 @@ Stimulant
 	Moderate Level.
 
 Bonus
-	The body generates Ephedrine.
+	The body generates Morphine.
 
 //////////////////////////////////////
 */
@@ -30,9 +30,9 @@ Bonus
 		var/mob/living/M = A.affected_mob
 		switch(A.stage)
 			if(5)
-				if (M.reagents.get_reagent_amount("ephedrine") < 10)
-					M.reagents.add_reagent("ephedrine", 10)
+				if (M.reagents.get_reagent_amount("morphine") < 15)
+					M.reagents.add_reagent("morphine", 15)
 			else
 				if(prob(SYMPTOM_ACTIVATION_PROB * 5))
-					M << "<span class='notice'>[pick("You feel restless.", "You feel like running laps around the station.")]</span>"
+					M << "<span class='notice'>[pick("You feel light on your feet.", "You feel weightless and free of obstructions.")]</span>"
 	return
--- a/code/datums/diseases/advance/symptoms/stimulant.dm
+++ b/code/datums/diseases/advance/symptoms/stimulant.dm
@@ -30,8 +30,8 @@ Bonus
 		var/mob/living/M = A.affected_mob
 		switch(A.stage)
 			if(5)
-				if (M.reagents.get_reagent_amount("morphine") < 15)
-					M.reagents.add_reagent("morphine", 15)
+				if (M.reagents.get_reagent_amount("morphine") < 9)
+					M.reagents.add_reagent("morphine", 9)
 			else
 				if(prob(SYMPTOM_ACTIVATION_PROB * 5))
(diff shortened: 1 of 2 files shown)


Swagile:
If there is any issues with the git, I am sorry. I will get right on it since I just got it working but from what I can tell, this is solid git wise.

GuyonBroadway:
Make it add coffee too in order to make it shake off coldness and wake you up and I think it'll be pretty good.

It would be a nice counter to lings and slings without making you invincible vs them.

Swagile:
Coldness? Barely any lings use cryosting and the moment a ling does that, people yell at the top of their lungs about it. Even then, its a legit ling strategy that is very hard to do since they have to switch between stings a lot in order to do it and the payoff is miniscule since the victim can keep fighting (aka throw rods, extinguishers, etc.)

Waking you up? Most lings don't even use any stings based on that. They stun with water / stun prod / stun anything; very few people actually buy the paralysis sting and I do not think coffee actually helps against that.

GuyonBroadway:
It'll also get people up from being stunned by baseballs faster IIRC, also lings DO use their cryosting fairly often, it;s one of ther more potent level one abilities, especially vs hardsuit users where the cold damage is more effective.

Plus it makes more sense for a "stimulants" virus to contain a stimulant like caffeine.

GuyonBroadway:
Not to mention it would allow people to survive in colder areas for a little longer.

Swagile:
Ill add this in a bit. I gotta go to sleep due to classes tomorrow. When I come back, ill add in coffee.

McBawbaggings:
Don't really like this idea and the current implementation is flawed:

Viruses tick every game tick and have a chance to go off.
SYMPTOM_ACTIVATION_PROB = 3 is the default chance used.
if(prob(SYMPTOM_ACTIVATION_PROB * 10)) == 30%.

The addiction value of morphine is 25.
The check value is <15
You add 15 each time this 30% chance goes.

So when you get the first dose, you have a 30% chance to start dying on the next roll.
You lose 0.8 morphine a tick so before you get below 11, you have 5 30% rolls to survive.
I think that's like an 80% chance to start overdosing(?)

Change the threshold to 10 if the addiction value is an exclusive value or 9 if it's an inclusive value.

And to why I think this is bad, virology is this pretty old system and it does need an overhaul to make it less tedious and grindy. But just changing the stimulant symptom which is a good legacy symptom and reminds me a lot of the trekchem chemists with hyperzine which was fun. Morphine is more of a powergaming chem since it removes slowdown.

McBawbaggings:
Actually scratch that, it'd be 4 rolls since 15 < 15 is 0.
Thus the chances of dying are 76%(?)
Even then, if you don't die, when you dip below 15, you still have that 76% chance to try every time.

McBawbaggings:
Shit, this is addiction, not overdose but still.

Swagile:
Do you want me to lower the amount then?

If not, I can go with the shitty version of my nerf which keeps Epherdine but removes the liquid plasma from Viro storage and makes Stim virus a level 5 virus instead.

Its a hacky way to do it, but if you can get addicted too fast on morphine with this even with lowered values, then thats the only other way I know of nerfing the stim virus.

McBawbaggings:
Addiction rate is inclusive. Change the check to 9.

JohnGinnane:
Code looks good

Swagile:
I made both 9 since even if morphine goes below 9 with Bawb's suggestion, it still adds 15 reagent in the code. So I changed both values to 9 for the time being. Unless I did my math wrong, it should keep the morphine at a constant 9.

McBawbaggings:
Yeah on retrospect, this counteracts anti stuns chems so is okay +1

Hackole:
I'm fine with this. +1.

Gandalf2k15:
Headmin approval gained (2/4). Merging.

PR #3044, opened 2016-11-16
More realistic hallucinations

Kierany9:
The max health hallucination now makes your stamina appear full as well. This also prevents you from gauging your HP from your stamina bar while using Morphine, Heroin or Miner's Salve.


McBawbaggings:
+1 Add changelog.

Hackole:
Was unaware that this was a thing you could do, get the log in like bawb said and then it'll be a +1

ArcLumin:
Did you remove the damage vignette? You can tell it from that too

ArcLumin:
Code checks out, merging

PR #3149, opened 2016-12-24
Fixes Book of Darkness spells not carrying over with Lichdom

### LINKED ISSUE #3147: Lichdom doesn't restore Book of Darkness spells
It does when sold separately, marking this down while I work on it.

TheOneAndOnlyCreeperJoe:
Fixes #3147 

The spells were tied to the mob not to the mind, which is ironic how it's a necromancy set and it's all about binding the mind to the bones and all that shit.


ArcLumin:
Code's good, merging as it's a bugfix

PR #3154, opened 2016-12-25
Standard holoparasite now has timestop + armor piercing + r_wall breaking

TheOneAndOnlyCreeperJoe:
What it says on the tin, standard holoparas now have timestop and give immunity to it to their owner (which is done by making a new timestop spell that is a child to timestop that sends a return value when cast), as well as being capable of breaking down reinforced walls.

Really needs a code review since there's probably a better way to go about this.

### DIFF
--- a/code/modules/mob/living/simple_animal/guardian/guardian.dm
+++ b/code/modules/mob/living/simple_animal/guardian/guardian.dm
@@ -300,7 +300,7 @@
 		playsound(loc, src.attack_sound, 50, 1, 1)
 		playsound(loc, src.attack_sound, 50, 1, 1)
 		playsound(loc, src.attack_sound, 50, 1, 1)
-	if(istype(target, turf/simulated/wall/r_wall))
+	if(istype(target, /turf/simulated/wall/r_wall))
 		var/turf/simulated/wall/r_wall/RW = target
 		src.say("[src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry]\
 			[src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry]")
--- a/code/modules/mob/living/simple_animal/guardian/guardian.dm
+++ b/code/modules/mob/living/simple_animal/guardian/guardian.dm
@@ -258,9 +258,10 @@
 /mob/living/simple_animal/hostile/guardian/punch
 	melee_damage_lower = 20
 	melee_damage_upper = 20
+	armour_penetration = 75
 	damage_transfer = 0.5
 	playstyle_string = "As a standard type you have the ability to stop time for everyone but you and your summoner, but have a high damage resistance and a powerful attack capable of smashing through walls."
-	environment_smash = 3
+	environment_smash = 2
 	magic_fluff_string = "..And draw the Assistant, faceless and generic, but never to be underestimated."
 	tech_fluff_string = "Boot sequence complete. Standard combat modules loaded. Holoparasite swarm online."
 	bio_fluff_string = "Your scarab swarm stirs to life, ready to tear apart your enemies."
--- a/code/modules/mob/living/simple_animal/guardian/guardian.dm
+++ b/code/modules/mob/living/simple_animal/guardian/guardian.dm
@@ -269,19 +269,19 @@
 
 /mob/living/simple_animal/hostile/guardian/punch/New()
 	..()
-	var/obj/effect/proc_holder/spell/aoe_turf/conjure/timestop/holoparasite/holoparasite = null
+	var/obj/effect/proc_holder/spell/aoe_turf/conjure/timestop/holoparasite/hg = null
 
-	holoparasite = new /obj/effect/proc_holder/spell/aoe_turf/conjure/timestop/holoparasite
-	holoparasite.clothes_req = 0
-	holoparasite.human_req = 0
-	holoparasite.player_lock = 0
-	AddSpell(holoparasite)
+	hg = new /obj/effect/proc_holder/spell/aoe_turf/conjure/timestop/holoparasite
+	hg.clothes_req = 0
+	hg.human_req = 0
+	hg.player_lock = 0
+	AddSpell(hg)
 
-	var/obj/effect/proc_holder/spell/self/timestopimmunity/timestopimmunity = null
+	var/obj/effect/proc_holder/spell/self/timestopimmunity/tsi = null
 
-	timestopimmunity = new /obj/effect/proc_holder/spell/self/timestopimmunity
+	tsi = new /obj/effect/proc_holder/spell/self/timestopimmunity
 
-	AddSpell(timestopimmunity) // I don't know how the hell it comes that the Holoparasite needs both timestop and the immunity to work, as humans can just have either and still walk around in it.
+	AddSpell(tsi) // I don't know how the hell it comes that the Holoparasite needs both timestop and the immunity to work, as humans can just have either and still walk around in it.
 
 /mob/living/simple_animal/hostile/guardian/punch/verb/Battlecry()
 	set name = "Set Battlecry"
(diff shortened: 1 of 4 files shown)


ghost:
Is this just regular wizman timestop that lasts for 9 seconds? If so, that's kinda waaaay too long for the standard. You can easily kill a cunt like that.

TheOneAndOnlyCreeperJoe:
aww shiet I didn't add the last commit before making the PR DNM

ghost:
Will the holoparasite shout "TOKI WO TOMARE" too?

Pyko1:
rwall smashing sounds really bad tho, basically means you can just walk in everywhere, with no effort.

Because that means you can pretty much get captains spare and rob armory in 1 minutes to round.
and kill ai at 2 minute mark.

At least I see many of the things are balanced around rwalls taking at least some effort to break.

TheOneAndOnlyCreeperJoe:
we're talking about holoparas thou, the guys who belong to traitors who can just buy c4 to bypass rwalls in a matter of seconds anyway.

also mcbaw yes, they do shout it.

Pyko1:
yeah but holoparas are already strong, and and invidual C4's cost TC and take a longer than just 1 punch. This guy can just walk right through station with no efffort.

ghost:
And genetics hulk which is easy to get makes you unstunnable while being able to smash walls at the cost of NO TC and being unable to use guns.

Pyko1:
With C4 you plant second or 2 and then there is 10 sec timer.
While with this you can manifest punch and return in less than a second.

Pyko1:
Hulks take effort to get and I recall they have just a chance of smashing walls down.

ghost:
Genetics doesn't take effort, force feed a monkey an unstable mutagen cig for a few ticks to get all the random mutations to narrow it down quickly.

TheOneAndOnlyCreeperJoe:
ok you can remove DNM

trifornt-tinfornt:
I'd rather they have a bit of armor penetration over being able to _instantly_ punch down rwalls, considering that riot armor would still stop them dead in their tracks. Also it's just a straight upgrade to hulk that way, and that wouldn't be as interesting. At least make it take multiple punches or something
However, this is a good start on buffing Standard.

Pyko1:
Could be interesting take on breaking rwalls if they had like 50% chance of taking 1 step of rwall deconstruction every punch.

So it would still be fairly fast but would have visual indicator as the wall would get "deconstructed" after every successful punch. And the hulks should also use same shit really.

TheOneAndOnlyCreeperJoe:
swapped out rwallsmash for armor piercing, percentage is 75%

yobob591:
I like pykos idea, you gotta ora ora the wall down not just one hit

gumner:
on my point of view,if you buy something that costs 20tc,you expect something good,or else just buy a .357 and a fuckton of ammo and kill the entire station without even using all your tc's

yobob591:
Should make it so you fly back after time unfreezes if you punch someone enough to knock them down during it, not doing any extra damage, just to toss them away

blobbernaut:
Change it so it has like, a 10% chance to break down the Rwall, like a durand.

TheOneAndOnlyCreeperJoe:
k it can fuck up r_walls now like a hulk.

yobob591:
i likey

ghost:
I don't think you can just override the attack function like that since you're changing the parameters of that function?

blobbernaut:
wait, will it work if you call ..() before calling the chance to break the wall? 

That should probably be after.

TheOneAndOnlyCreeperJoe:
someone point out why it won't work while I take a nap because I can't see it.

Pyko1:
What it says when you compile?

code/modules/mob/living/simple_animal/guardian/guardian.dm
```diff
@@ -258,14 +258,31 @@
...
+	holoparasite.clothes_req = 0
+	holoparasite.human_req = 0
+	holoparasite.player_lock = 0
+	AddSpell(holoparasite)
+
+	var/obj/effect/proc_holder/spell/self/timestopimmunity/timestopimmunity = null
+
+	timestopimmunity = new /obj/effect/proc_holder/spell/self/timestopimmunity
+
+	AddSpell(timestopimmunity) // I don't know how the hell it comes that the Holoparasite needs both timestop and the immunity to work, as humans can just have either and still walk around in it.
```
JohnGinnane: You might be able to get around this using the `immune` list defined in the `effect/timestop`:

https://github.com/HippieStationCode/HippieStation13/blob/56a3fdddeb58b6e0dedb20514f4cc43bee628c58/code/modules/research/xenobiology/xenobiology.dm#L549

Here you can see it skips people who are in the list. If you add summoner to this list then when the para uses it both the para and summoner should be immune.

Here's where the caster is added (usually the wizard):

https://github.com/HippieStationCode/HippieStation13/blob/56a3fdddeb58b6e0dedb20514f4cc43bee628c58/code/modules/research/xenobiology/xenobiology.dm#L538

code/modules/mob/living/simple_animal/guardian/guardian.dm
```diff
@@ -258,14 +258,31 @@
...
+	playstyle_string = "As a standard type you have the ability to stop time for everyone but you and your summoner, but have a high damage resistance and a powerful attack capable of smashing through walls."
 	environment_smash = 2
 	magic_fluff_string = "..And draw the Assistant, faceless and generic, but never to be underestimated."
 	tech_fluff_string = "Boot sequence complete. Standard combat modules loaded. Holoparasite swarm online."
 	bio_fluff_string = "Your scarab swarm stirs to life, ready to tear apart your enemies."
 	var/battlecry = "AT"
 
+/mob/living/simple_animal/hostile/guardian/punch/New()
+	..()
+	var/obj/effect/proc_holder/spell/aoe_turf/conjure/timestop/holoparasite/holoparasite = null
```
JohnGinnane: From what I've seen we usually name variables like this "HP" or something, instead of just "type" as it's very confusing right now. Especially below when you go "`holoparasite.clothes_req..`"

code/modules/mob/living/simple_animal/guardian/guardian.dm
```diff
@@ -285,6 +300,17 @@
 		playsound(loc, src.attack_sound, 50, 1, 1)
 		playsound(loc, src.attack_sound, 50, 1, 1)
 		playsound(loc, src.attack_sound, 50, 1, 1)
+	if(istype(target, /turf/simulated/wall/r_wall))
+		if(prob(10))
+			playsound(src, 'sound/effects/meteorimpact.ogg', 100, 1)
+			src << text("<span class='notice'>You smash through the wall.</span>")
+			src.say("[src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry][src.battlecry]\
```
JohnGinnane: I think you can move the say to outside of the if..else because it'll happen either way. And what about making the meteor noise in both cases, much like a hulk smashing a wall?

code/modules/research/xenobiology/xenobiology.dm
```diff
@@ -538,6 +537,8 @@
 	for(var/mob/living/M in player_list)
 		for(var/obj/effect/proc_holder/spell/aoe_turf/conjure/timestop/T in M.mind.spell_list) //People who can stop time are immune to timestop
 			immune |= M
+		for(var/obj/effect/proc_holder/spell/self/timestopimmunity/T in M.mind.spell_list) //Used by stand users mainly so that their own holopara doesn't timestop them.
```
JohnGinnane, line 540: Oh I see you're working with it here. Yeah I would say just manually add summoner and para to immune list when you create the para

code/modules/research/xenobiology/xenobiology.dm
```diff
@@ -546,26 +547,31 @@
...
 			for(var/mob/living/M in orange (freezerange, src.loc))
-				if(M in immune)
-					continue
-				M.stunned = 10
-				M.anchored = 1
-				if(istype(M, /mob/living/simple_animal/hostile))
-					var/mob/living/simple_animal/hostile/H = M
-					H.AIStatus = AI_OFF
-					H.LoseTarget()
+				if(!istype(M, /mob/living/simple_animal/hostile/guardian/punch))
```
JohnGinnane: This is suuuuper snowflakey, pls no((

JohnGinnane (changes requested):
Might need to revise how the immunity works, not particularly happen with snowflake code

code/modules/research/xenobiology/xenobiology.dm
TheOneAndOnlyCreeperJoe: no other way, because the timestop effect ends hostile mob AI, under which the guardian falls. Even giving them timestop is not enough to circumvent this.

TheOneAndOnlyCreeperJoe:
@Pyko1 code/modules/mob/living/simple_animal/guardian/guardian.dm:309:error: target.dismantle_wall: undefined proc

TheOneAndOnlyCreeperJoe:
It works but the spam filter will automute them for saying their battlecry too often, can't do anything about that.

ghost:
I didn't say +1 before.
+1

JohnGinnane:
Is this ready for codecheck?

yobob591:
Hackole the time stop is the core of this buff tho, I can understand removing the rwall shit but standard and timestop are meant for eachother. Maybe nerfing the duration if it comes to it

ghost:
The duration is 3.5 seconds which is pretty short.

Swagile:
holoparasite is 20 fucking TC, it defo needs a buff besides giving you extra HP (which doesn't matter as much as you think it does because lol stun combat) and a bigger punch (who the fuck cares when you can dodge it and laser standard to death?)

timestop negates some of its weaknesses while not making it too OP, plus everyone can see timestop so they know you have a standard holo and can find ways around timestop but still makes it effective

blobbernaut:
Standard has a lowish range and only does 30 damage with no armor penetration. It's hard as fuck to actually hit somebody who is running away. This is a necessary buff, hackole.

Swagile:
ill wait till requested changes / code is fully OK'd before giving my GM vote

TheOneAndOnlyCreeperJoe:
Screw the requested changes, there's no better way of doing this to my knowledge due to hardcoding.

Snowflake code is needed because the holoparasite is considered for all intents and purposes a hostile mob and freezing the AI, even when it's player controlled, stuns the holoparasite.

Swagile:
yeah if there isn't a way to fix it and john can confirm, then ill gm +1 this

TheOneAndOnlyCreeperJoe:
Too uninterested in maintaining this PR because Timestop code is too large of a fucking cunt to work with, someone can do it themselves if they reaaally want it.

Swagile:
rip

yobob591:
ill be taking it from ere

Issue #3039, opened 2016-11-15
Roman Shield was unaffected by shield nerfs


ArcLumin:
MENTOR: Hackole: oh yeah arc i'm a lazy butthole can you make a pr for me

The shield nerfs did not affect the roman shield, meaning you can easily get a better shield than what rnd and sec have

ghost:
GEE THANKS CRYSTAL

PR #3162, opened 2016-12-26
Nerfed the Roman Shield because it was better than any of the security shields

Spacedong:
Fixes issue #3039

fuck you crystal

### DIFF
--- a/code/game/objects/items/weapons/shields.dm
+++ b/code/game/objects/items/weapons/shields.dm
@@ -19,7 +19,7 @@
 	w_class = 4
 	materials = list(MAT_GLASS=7500, MAT_METAL=1000)
 	origin_tech = "materials=2"
-	block_chance = list(melee = 60, bullet = 50, laser = 50, energy = 50) //Skeleton-exclusive atm, sooo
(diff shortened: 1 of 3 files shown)


ArcLumin:
Minor reduction, still no slowdown or swap, I guess I'm fine with this.

McBawbaggings:
+1

Hackole:
I'd prefer it be kept as is and just removed from the autodrobe, honestly.

GuyonBroadway:
Can just make a child of the roman shield called replica roman sheild and set all its resistancies to be 0, add that to the autodrobe so clowns can pretend to join the chaplain crusade. That way skeleton thralls from the necromancer stone don't get gimped by proxy

yobob591:
that sounds way better

blobbernaut:
should just give a shitty version to the autodrobe, skeletons should be stronk.

Pyko1:
If we make replica Roman shield, it still should have at least some resistance since there is only 1 and it costs coin.

Spacedong:
ok

PR #3168, opened 2016-12-27
Stand Arrow

yobob591:
Adds a stand arrow to space which acts as an infinite holoparasite injector with the downside of having a 40% chance of killing the person who it is used on. It is also able to be used on others. Obviously if they die its on you unless they asked for it.


yobob591:
wew that was a lot of codin- jesus christ its already 7 pm

yobob591:
fucking really

Issue #3236, opened 2017-01-01
Meatspiking humans does not check for the 'occupied' var


Spacedong:
will fix in a new PR soon, just wanted to have a heads up

Spacedong:
this just means that you can meatspike humans while a mob is on the spike

McBawbaggings:
DONG DONG NEVER DIE

PR #3237, opened 2017-01-01
Implemented a check for the occupied var when meatspiking humans while a mob is already on the spike

Spacedong:
Prevents slamming a human on top of a monkey/xeno on a meatspike. Fixes issue #3236.


McBawbaggings:
Okay, merging.

PR #3265, opened 2017-01-03
Gang Tweaks, Traitorling Playercount and Maximum Playercounts

Kierany9:
### Adds maximum reccomended playercounts:

Admins can still force these modes above the maximum, they simply won't appear in secret. This is intended for modes that don't run well enough in highpop, as they don't pose enough of a threat to the station. **Abductors now have a maximum playercount of 45** and they have been given a special scaling to acomodate this.

### Changes minimum playercounts for certain modes.

 - Traitorling: 0 -> 25
 - Gang: 20 -> 25

### Adjusts the number of gangs to appear in any given gang round.

The minimum number of gangs has been increased from 2 to 3. Extra gangs now have a chance equal to the number of players to appear at the 38 and 50 player marks. The maximum number of gangs is now five. This is why I've slightly bumped the minimum playercount. Because having two gangs and one getting fucked roundstart is never fun.

### Nerfs the Rigatoni by increasing the price of some of their items.

>let's give the rigatoni a c20 for less price than an uzi amirite oh and how about a detvolver at the price of a pistol
 - Tommy Gun: 55 -> 80 Influence
 - Tommy Gun Ammo: 35 -> 50 Influence
 - Revolver: 25 -> 30 Influence
 - Revolver Ammo: 10 -> 20 Influence

For reference, the Uzi and Sketckin are the other gang equivalents and are worth 60 and 25 influence respectively, and 40 and 10 for the ammunition. The Tommy Gun deals 10 less damage per bullet than the Uzi, but a single bullet deals enough stamina damage to incapacitate. Revolvers deal 15 less damage than the pistol but deal very nice stun and stamina damage.

### Recalling now costs 25 influence and is doubled for every successful recall.

Because this shit was annoying with three gangs, let alone five.

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -670,6 +670,6 @@
 		if(probabilities[M.config_tag]<=0)
 			qdel(M)
 			continue
-		if(M.required_players <= crew)
+		if(M.required_players <= crew && (!M.max_players || M.max_players >= crew))
 			runnable_modes[M] = probabilities[M.config_tag]
 	return runnable_modes
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -655,7 +655,7 @@
 		if(probabilities[M.config_tag]<=0)
 			qdel(M)
 			continue
-		if(M.can_start())
+		if(M.can_start() == 1)
 			runnable_modes[M] = probabilities[M.config_tag]
 			//world << "DEBUG: runnable_mode\[[runnable_modes.len]\] = [M.config_tag]"
 	return runnable_modes
--- a/code/game/gamemodes/abduction/abduction.dm
+++ b/code/game/gamemodes/abduction/abduction.dm
@@ -25,8 +25,7 @@
 	world << "<b>Crew</b> - don't get abducted and stop the abductors."
 
 /datum/game_mode/abduction/pre_setup()
-	// abductor_teams = max(1, min(max_teams,round(num_players()/config.abductor_scaling_coeff))) Old scaling
-	abductor_teams = max(1, min(max_teams,round(max_teams*num_players()/max_players))) // Scaling based off of max players
+	abductor_teams = max(1, min(max_teams,round(num_players()/config.abductor_scaling_coeff)))
 	var/possible_teams = max(1,round(antag_candidates.len / 2))
 	abductor_teams = min(abductor_teams,possible_teams)
 
--- a/code/game/gamemodes/abduction/abduction.dm
+++ b/code/game/gamemodes/abduction/abduction.dm
@@ -9,6 +9,7 @@
 	antag_flag = ROLE_ABDUCTOR
 	recommended_enemies = 2
 	required_players = 15
+	max_players = 50
 	var/max_teams = 4
 	abductor_teams = 1
 	var/list/datum/mind/scientists = list()
@@ -24,7 +25,8 @@
 	world << "<b>Crew</b> - don't get abducted and stop the abductors."
 
 /datum/game_mode/abduction/pre_setup()
-	abductor_teams = max(1, min(max_teams,round(num_players()/config.abductor_scaling_coeff)))
+	// abductor_teams = max(1, min(max_teams,round(num_players()/config.abductor_scaling_coeff))) Old scaling
+	abductor_teams = max(1, min(max_teams,round(max_teams*num_players()/max_players))) // Scaling based off of max players
 	var/possible_teams = max(1,round(antag_candidates.len / 2))
 	abductor_teams = min(abductor_teams,possible_teams)
 
--- a/code/game/gamemodes/abduction/abduction.dm
+++ b/code/game/gamemodes/abduction/abduction.dm
@@ -9,7 +9,7 @@
 	antag_flag = ROLE_ABDUCTOR
 	recommended_enemies = 2
 	required_players = 15
-	max_players = 50
+	max_players = 45
 	var/max_teams = 4
 	abductor_teams = 1
 	var/list/datum/mind/scientists = list()
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -55,7 +55,7 @@
 		if(playerC < required_players)
 			return 0
 		if(max_players && (playerC > max_players))
-			return 0
+			return 2
 	antag_candidates = get_players_for_role(antag_flag)
 	if(!Debug2)
 		if(antag_candidates.len < required_enemies)
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -54,6 +54,8 @@
 	if(!Debug2)
 		if(playerC < required_players)
 			return 0
+		if(max_players && (playerC > max_players))
+			return 0
 	antag_candidates = get_players_for_role(antag_flag)
 	if(!Debug2)
 		if(antag_candidates.len < required_enemies)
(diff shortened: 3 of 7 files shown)


McBawbaggings:
Good changes, I think abductor could maybe have max pop at 40.
Otherwise +1.

trifornt-tinfornt:
If you're adding fucking five gangbosses, for the love of God, remove their ability to recall the shuttle from anywhere at the press of a button, the same button that's right next to "Buy An Uzi".
2 hour long gang rounds that don't go anywhere because the five people that are alive alive are hiding, the shuttle keeps getting recalled, and everyone else is dead are extremely monotonous.
I like the idea of more gangbosses (besides the fact that it's going to make sec get run over even faster) but this can't just be a single number that's changed and left.

oprayx77:
>nerfing my rigatoni masterrace

yeah you should go ahead with that :)

Dudeinyourbackyard:
heck yeah but iunno about having five gangs at once, seems pretty crazy even for high pop.
and yeah abductors' max should be lowered to 40, 50 is still pretty high for such a relatively quiet gamemode.

Kierany9:
# Recalling now costs 25 influence, and doubles every time it is used.

Kierany9:
Abductors are at 45 instead of 50 players maximum. Max players actually works now.

JohnGinnane:
These are wonderful changes

code/game/gamemodes/abduction/abduction.dm
```diff
@@ -24,7 +25,8 @@
 	world << "<b>Crew</b> - don't get abducted and stop the abductors."
 
 /datum/game_mode/abduction/pre_setup()
-	abductor_teams = max(1, min(max_teams,round(num_players()/config.abductor_scaling_coeff)))
+	// abductor_teams = max(1, min(max_teams,round(num_players()/config.abductor_scaling_coeff))) Old scaling
+	abductor_teams = max(1, min(max_teams,round(max_teams*num_players()/max_players))) // Scaling based off of max players
```
JohnGinnane: Not based on config at all? Shouldn't we do that?

code/controllers/configuration.dm
```diff
@@ -655,7 +655,7 @@
 		if(probabilities[M.config_tag]<=0)
 			qdel(M)
 			continue
-		if(M.can_start())
+		if(M.can_start() == 1)
```
JohnGinnane: why?

code/game/gamemodes/game_mode.dm
```diff
@@ -53,6 +54,8 @@
 	if(!Debug2)
 		if(playerC < required_players)
 			return 0
+		if(max_players && (playerC > max_players))
+			return 2
```
JohnGinnane: What does 2 signify?

JohnGinnane:
Have you tested this locally?

KittyKathy:
In regards to recalling the shuttle what happens from the choices below when someone recalls the shuttle using their gang tool?

1. The cost to recall only doubles for that individual gang member that recalled the shuttle.
2. The cost to recall doubles for the whole gangs whom the individual is in that recalled the shuttle. 
An example would be if one of the Gene gang leaders/lieutenants recalled the shuttle the next time a gene gang leader/lieutenants recalls the shuttle they pay double influence even if it was not the same person recalling. (This is so you can't have them take turns recalling it at base influence cost of 25 which can be like 3 times before they start paying double influence cost).
3. The cost to recall doubles for every gang leaders/lieutenants when any of them recalls the shuttle. 
(This is so the unlikely event of each gang taking turns to recall the shuttle at base cost does not happen.

It would probably be better if they was just a set amount of times it could be recalled a round like maybe 3 and/or they can not recall the shuttle after a certain time has past since the shift started like say 1 hour? I mean if they failed to secure the station by the one hour mark what were they doing? They probably got wrecked and the last leader/lieutenant is hiding/fortifying somewhere with a billion reinforced walls to dominate or the gangs are too busy murderboning and just wants to extend the round as much as possible.

Also to note is that they can dominate to stop the shuttle from leaving why do they need the ability to recall from anywhere when they can just use a communications console with head level access, do we really need to make it easy for them to extend the round as much as possible until admins have to do something?

JohnGinnane:
Good point about being able to freely recall when you have the access (mid to late game, depending on robustness). 

I like the idea of limiting the amount of recalls but this means it can be called/recalled by heads to quickly end the round against gangs (which might even be a valid option?).

Kierany9:
Options 1 and 2 are the same, as only the boss can recall the shuttle, lieutenants cannot. As for the reason why they can recall in the first place, I assume it's to prevent heads/sec securing the comms board at the first sign of gang activity and immediately calling the shuttle. Either way, if recalling is still an issue in the future, it can be changed further.

code/game/gamemodes/abduction/abduction.dm
Kierany9: It probably should be done that way but linear scaling with the config setting is pretty ass for this particular mode.

code/controllers/configuration.dm
Kierany9: because 2

code/game/gamemodes/game_mode.dm
Kierany9: Signifies that it should not start unless forced by an admin, hence the above == 1 for the standard secret gamemode selection

code/game/gamemodes/abduction/abduction.dm
JohnGinnane: Well we could put max teams and max players into the config file, and then you can set those vars by reading from the file?

Kierany9:
@JohnGinnane Yes it has been tested locally and works.

code/game/gamemodes/game_mode.dm
JohnGinnane: Oh I see, cool

JohnGinnane:
Neat

code/game/gamemodes/abduction/abduction.dm
Kierany9: On second thought the old scaling isn't too bad, 8 antags in a 45 player round isn't the most sound of ideas.

It's also because the thought of touching config files makes me want to eat rocks.

Kierany9:
merge dis

KittyKathy:
Oh so only one person from each gang can recall using the gang tool which is the original boss?
How many bosses that can recall the shuttle spawns in each gang? I assume one which is the original boss.

Kierany9:
yes

KittyKathy:
Okay and just to make sure I understand, every time recall is used it doubles the cost for the next person to use it regardless of which gang they are in?
So if Max gang recalls it then gene gang recalls it a short while after they pay double influence to recall?
If it works like that should be good to stop mass recalling at least using the gang tool.
Good PR maybe you can make the sleeping carp gang balanced with the rest of the gangs next as guns > melee.

meteoses:
Hmm. Now, I know the "git gud" argument isn't appreciated by most; but in this case it really is a matter of "git gud". Think of how many Gang rounds you've watched me take over as Captain; if you mobilize your troops, quickly arm Sec, implant the lawyers/others who show up to Sec, then move with 8+ people to cargo, then have your forces go out to stun+cuff anyone in the area then bring them back to cargo where you are ordering implants, tell them to stick together => free win.

It's really the most controlled game mode, you can almost always strangle them out if you are Captain/HoP/HoS. They don't need nerfs. The other changes are good.

Kierany9:
>Okay and just to make sure I understand, every time recall is used it doubles the cost for the next person to use it regardless of which gang they are in?

No, the recall cost only doubles for each individual gang boss.

>Meteoses

The nerf isn't because Rigatoni is OP against Sec, it's because they're OP compared to other gangs. Not only is their gear better than standard equipment, it's cheaper too. Besides, the extra gangs should bring sec in line.

MethMedic:
Yea the rigatoni guns were insanely good

JohnGinnane:
Heads are happy, merging

PR #3277, opened 2017-01-04
Fixes magic runtime

ArcLumin:
Fixes the runtime with the staff of nothing trying to create a non-existent ammo casing.

Fixes https://github.com/HippieStationCode/HippieStation13/issues/3274


TDcoolguy300:
More fixes.

PR #3287, opened 2017-01-05
Fixes missing icon in hockey set.

GuyonBroadway:
In my fervour to fix what I am sure Spacedong broke with a saxophone I missed the inhand icons for the stick. This fixes that.


ArcLumin:
Merging as it's a fix

Issue #3123, opened 2016-12-21
Gas canisters once wrenched to connectors are still visibly wrenched down even if the connector they were on does not exist anymore


Spacedong:
title, 

example
https://vgy.me/2ituLe.png

Spacedong:
i fixed this

PR #3298, opened 2017-01-06
Canisters wrenched to connectors will now update their icons properly if the connector they were attached to is deleted.

Spacedong:
Fixes issue #3123.


JohnGinnane:
Code looks good, surprising how simple a fix that was. Merging

PR #3313, opened 2017-01-07
Fixes the lag exploit for real by adding a fail safe.

TheOneAndOnlyCreeperJoe:
Told you @McBawbaggings that it wasn't the sparks.

Adds a failsafe that stops a single item from repeatedly bouncing against a containment field wall, triggering several failsafes.


ArcLumin:
Code looks good, merging.

PR #3316, opened 2017-01-08
respect my pronouns

Spacedong:
which retard fucked this up


JohnGinnane:
Code is good, merging as it's bug fixes/typo and don't need head input

PR #3318, opened 2017-01-08
Fixes #2367 and tweaks galaxy superfart

### LINKED ISSUE #2367: you butt flies off into the galaxy broken
it's supposed to make an immoveable butt rod that cuts through the station in a random direction from you but it doesn't actually do this it just gibs you

### LINKED ISSUE #3088: Immovable butt doesn't work
There should be a .4% chance to superfart an immovable rod. Sadly, you only get the message that your butt flies off into the galaxy and nothing to accompany it.

KittyKathy:
:cl: Kitty Kathy
   bugfix: Galaxy superfart now works, but works a bit different that the original code.
   tweak: Galaxy superfart does not shoot off into a random direction it goes opposite what you are facing.
   tweak: Galaxy superfart will also move in a straight line until it reaches the end of the map.
   tweak: Immovable rods and their children now bump into whatever they spawn on.
   tweak: You can not galaxy superfart on any zlevel other than 1, it will instead go supernova if it picks that.
   /:cl:
Fixes #2367.
Am i doing this automatic changelog thing right?
I have no idea what I am doing to be honest never used git before.
Yes I tested this and it works unless something went horribly wrong when i was fooling around with the command line trying to get this to work.
*Edited wrong issue number

### DIFF
--- a/code/modules/events/immovable_rod.dm
+++ b/code/modules/events/immovable_rod.dm
@@ -25,7 +25,7 @@ In my current plan for it, 'solid' will be defined as anything with density == 1
 	new /obj/effect/immovablerod(startT, endT)
 
 /obj/effect/immovablerod
-	name = "Immovable Rod"
+	name = "immovable rod"
 	desc = "What the fuck is that?"
 	icon = 'icons/obj/objects.dmi'
 	icon_state = "immrod"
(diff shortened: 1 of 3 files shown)


Pyko1:
`H.visible_message("<span class='danger'>[H.name] is penetrated by an immovable [name]!</span>"`

Mc Grifins is penetrated by an immovable immovable rod!

lcass:
once pyko's thingy is fixed it looks alright. Dunno why this wasn't doesn't a while ago.

JohnGinnane:
Code looks good. The way we do changelogs is by creating a .yml file in /html/changelogs/

Have a look at example.yml in that folder to see how it's done

ArcLumin:
Looks good to me, would consider this to be a bugfix so it's good to merge once it gets an actual changelog

ArcLumin:
Changelog is good, merging as it's a fix

PR #3325, opened 2017-01-08
Growing icons for limb and butt flowers

tortellinitony:
They should have had them before, and now they do.


ghost:
noice

ArcLumin:
In my opinion this is a bugfix. Code's good, it's just icons. Merging as it's a bugfix

PR #3331, opened 2017-01-09
Heroin+Aranesp nerfs, readds Stimpacks

Kierany9:
- Stimpacks have been readded at their old price of 5 TC and no longer reduce stuns, but rather allow you to ignore slowdowns. Overdose has been made to have an actual consequence(if you were stupid enough to inject two stimpacks) and the speedboost only applies when you wouldn't normally be slowed down, IE, above 60 HP. Healing now stops at 60 instead of 50 HP.
- Heroin and Aranesp have reduced stamina regen, from full immunity to +15 per tick and made Aranesp's speedboost less game-breaking(FAST + VERY_FAST -> VERY_FAST). Aranesp has also been made slightly less toxic.

- ~Some changes have been made to bath salts incase I uncomment it in the near future. The antistun effects will only kick in upon overdose, which completely nullifies the reagent's metabolization for 45 ticks and then resumes it at half speed. Attempting to cheese this with charcoal/calomel deals serious toxin damage and renders any bonuses useless as well. Once it wears out, you will be paralyzed for 15 ticks as well as taking massive brain damage and toxin damage based on time spent in the system, The addictions have been made more crippling and will most likely culminate in death by screaming.~

- Ephedrine is now fast again. (At max HP)


MethMedic:
If there is no loophole to powergame, sure

Kierany9:
re-commenting bath salts because even though it's literally a suicide drug, i don't want to deal with the flak for the inevitable powergame despite how hard i've made it

KittyKathy:
Can you loophole the Stimulants needing full health for the speed boost by combining it with Aranesp to get the FAST speed boost and using something like doctors delight/omnizine and atropine to negate all the side effects of Arenesp?
Also because Stimulants come in a syringe can you not inject this into a vape and vape it to basically have unlimited Stimulants?
So now you can have people on Stimulants + Arenesp(including the counters for the side effects) for the whole round?
Also fix the description for Aranesp when?
For Bath salt you want chemicals that purge other chemicals to do toxin damage what about Pentetic Acid also no one ever uses Bath salts and probably never will.
There are easier ways to remove all chemicals from your body if that is what you are trying to prevent from happening when you have Bath salts in you, if you do a water + potassium reaction in your body using low ammounts like 1-5 unit each it clears your body of all chemicals similarly like a smoke reaction.
Bring back Anti-toxin when?

Kierany9:
>Can you loophole the Stimulants needing full health for the speed boost by combining it with Aranesp to get the FAST speed boost?

I mean, that's just using two chems at once to get different effects, like using Heroin and Aranesp, so it's not really a loophole.

>Also because Stimulants come in a syringe can you not inject this into a vape and vape it to basically have unlimited Stimulants?

fucking vapes stop being broken

> For Bath salt you want chemicals that purge other chemicals to do toxin damage what about Pentetic Acid

Oh right.

>also no one ever uses Bath salts and probably never will.

:^)

KittyKathy:
Yeah but isn't the whole point of the "you need full health for the speed boost" to prevent people from running away when injured at sanic speed? Well they still can if do they a lot a bit chemistry but i guess one stun is it for them so it does not matter.

Your bath salt code it says on line 511 it will do that if statement to do toxic damage if you have those reagents in you that also heal toxic damage but the statement on line 513 is a else if statement that means if the if statement on line 511 is true it will skip that else if right?
So what is to stop a person being stupid and from filling them self with large amounts of Charcoal, Pentetic Acid and Bath salts and being super stun immune without that whole side effect of paralyzing you if they keep refilling them self before they run out?

Fix Aranesp description when?Bring back Anti-toxin when?Your love for bath salts is weird.

KittyKathy:
Also Heroin is a shit tier drug just because it doubles stuns even morphine is better than that and it makes you sleep when you have less that 30 hp not even critical and makes you drowsy when it leaves your body.

KittyKathy:
Oh apparently I read this wrong and you were actually using the old way to have anti stuns thought that it was to stun you there but it was actually removing stuns why are you using the old way and not the new way you added? and why do those three chemicals do toxin damage? if it is so you can't heal the damage the drug does there are other reagents to heal the damage it does.

Kierany9:
>Fix Aranesp description when?

Already done, take a look

>Bring back Anti-toxin when?

Never

>Your love for bath salts is weird.

yes

Kierany9:
reverted salt changes because this isnt happening, slightly lessened the aranesp nerf and made ephedrine fast again

KittyKathy:
Fixed the description you say?

> Amps you up and gets you going, fixes all stamina damage you might have but can cause toxin and oxygen damage."

> fixes all stamina damage

> all

I don't know this seems to imply it fixes all stamina damage instantly maybe it's just me.

Bringing back sanic Ephedrine(almost no downside to runing like sanic) this surely won't end bad also the stimulants virus still makes morphine because grudge code logic.

KittyKathy:
What do you have against Anti-toxin? I will eventually try to bring it back myself if you won't.

banthebantz:
what why, aranesp did nothing wrong!

trifornt-tinfornt:
"made ephedrine fast again"
>completely forgot what made ephedrine so infuriating and decided to re-add it even though it's easy as piss to make
And you say you know what you're doing when it comes to balance!

GuyonBroadway:
You should make the stimpacks into medipen auto injectors for easy application.

SpankMasterson:
+1

Kierany9:
>traitor-only chem has speedboosts at max HP only

>ez to make chem available for all has unrestricted speed boosts

It's almost as if I'm retarded and these should be backwards

Kierany9:
Oh god these descriptions...

brb fixing

KittyKathy:
We can't even make Oxandrolone/Salicyclic Acid anymore why do we still have it in the code...is it used somewhere?

McBawbaggings:
+1

banthebantz:
Aranesp still did nothing wrong

blobbernaut:
ALWAYS NERF CHEMISTRY MAKE EVERY CHEMISTRY DRUG SHITTY

ArcLumin:
Code's good

lcass:
As long as the vape issue isn't actually an issue this should be fine , infini vape heroin does sound like a good time though.

blobbernaut:
Stimpacks should reduce stuns also readd adrenals

Kierany9:
merge dis

Pyko1:
had +2 and arc checked the code, merging

PR #3361, opened 2017-01-17
Ports Holy Explosions

asskek:
Ports holy water explosions from /tg/. Mixing 100u potassium and 100u holy water will result in an explosion that stuns+reveals nearby revenants and sets cultists on fire, searing their skin. Lots of little flavor things thrown in for good measure, hope you guys like it.

### DIFF
--- a/code/modules/reagents/Chemistry-Recipes/Pyrotechnics.dm
+++ b/code/modules/reagents/Chemistry-Recipes/Pyrotechnics.dm
@@ -60,7 +60,7 @@
 			R.stun(20)
 			R.reveal(100)
 		sleep(20)
-		for(var/mob/living/carbon/M in get_hearers_in_view(round(created_volume/48,1),get_turf(holder.my_atom)))
+		for(var/mob/living/carbon/M in get_hearers_in_view(round(created_volume/20,1),get_turf(holder.my_atom)))
 			if(iscultist(M))
 				M << "<span class='userdanger'>The divine explosion sears you!</span>"
 				M.Weaken(2)
--- a/code/modules/reagents/Chemistry-Recipes/Pyrotechnics.dm
+++ b/code/modules/reagents/Chemistry-Recipes/Pyrotechnics.dm
@@ -34,6 +34,41 @@
 	e.start()
 	holder.clear_reagents()
 
+/datum/chemical_reaction/reagent_explosion/potassium_explosion/holyboom
+	name = "Holy Explosion"
+	id = "holyboom"
+	required_reagents = list("holywater" = 1, "potassium" = 1)
+
+/datum/chemical_reaction/potassium_explosion/holyboom/on_reaction(datum/reagents/holder, created_volume)
+	var/atom/H = holder.my_atom
+	var/location = get_turf(holder.my_atom)
+	var/datum/effect_system/reagents_explosion/e = new()
+	world<<"created [created_volume] holyboom"
+	if(created_volume >= 150)
+		playsound(get_turf(holder.my_atom), 'sound/effects/pray.ogg', 80, 0, round(created_volume/48))
+		message_admins("A holy explosion has occurred at <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[H.x];Y=[H.y];Z=[H.z]'>([H.x],[H.y],[H.z])</a> last touched by [key_name_admin(H.fingerprintslast)]")
+		log_game("A holy explosion has occurred at <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[H.x];Y=[H.y];Z=[H.z]'>([H.x],[H.y],[H.z])</a> last touched by [key_name_admin(H.fingerprintslast)]")
+		e.set_up(round (created_volume/10, 1), location, 0, 0)
+		e.start()
+		holder.clear_reagents()
+		for(var/mob/living/simple_animal/revenant/R in get_hearers_in_view(7,get_turf(holder.my_atom)))
+			var/deity
+			if(!(deity_name == null))
+				deity = deity_name
+			else
+				deity = "Christ"
+			R << "<span class='userdanger'>The power of [deity] compels you, motherfucker!</span>"
+			R.stun(20)
+			R.reveal(100)
+		sleep(20)
+		for(var/mob/living/carbon/M in get_hearers_in_view(round(created_volume/48,1),get_turf(holder.my_atom)))
+			if(iscultist(M))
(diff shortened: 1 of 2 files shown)


yobob591:
👍

Anderbergst:
:+1:

asskek:
@spank @mcbawbaggings @dudeinyourbackyard input pls

PopNotes:
it's ok

gumner:
+1 anything that can counter cult is welcome

code/modules/reagents/Chemistry-Recipes/Pyrotechnics.dm
```diff
@@ -34,6 +34,41 @@
...
+/datum/chemical_reaction/reagent_explosion/potassium_explosion/holyboom
+	name = "Holy Explosion"
+	id = "holyboom"
+	required_reagents = list("holywater" = 1, "potassium" = 1)
+
+/datum/chemical_reaction/potassium_explosion/holyboom/on_reaction(datum/reagents/holder, created_volume)
+	var/atom/H = holder.my_atom
+	var/location = get_turf(holder.my_atom)
+	var/datum/effect_system/reagents_explosion/e = new()
+	world<<"created [created_volume] holyboom"
```
ghost: lmao

code/modules/reagents/Chemistry-Recipes/Pyrotechnics.dm
```diff
@@ -34,6 +34,41 @@
...
+	if(created_volume >= 150)
```
ghost: ?? Won't you need 75 and 75??

asskek: will fix when i get home, the /tg/ version has it as a lategame item you need bluespace beakers/grenades for.

code/modules/reagents/Chemistry-Recipes/Pyrotechnics.dm
asskek: yeah the world thing was for testing, will fix when i get home. my bad.

JohnGinnane:
I think we need to be aware of what this means. People can now antag check anyone they come across for cult by using this, which means cultists are getting seriously nerfed. This might be a little unfair..

asskek:
john: the original plan was to have it be 75-75, so you'd either need to make a grenade (wasting large beakers) or use bluespace beakers. not even sure if thos works in nades.

ArcLumin:
They work in nades just fine. Also so long as you have glass you can make large beakers

Pyko1:
Should cause some normal damage and extra to cult/revenant. So you cant just go around with largebeaker and couple beakers of potassium and a watertank full of holy water and explode it next to anyone you see.

Pyko1:
And cant really balance it with mix amount, as you could just fill chem tank with potassium and keep tossing needed amount of holywater in it every so often.

JohnGinnane:
Yeah no matter what, it means cultists are only antag that can be checked...

doobiedoo23:
this is retarded

KayeArray:
Why do you hate cult?

asskek:
This still causes a normal potassium explosion. If you go around using this willy-nilly you will be banned for randomly bombing people. This will, in practice, only be used to clear confirmed cult bases -- no one's going to go around bombing huge groups of people to test for cult.

asskek:
And why is it retarded, Doobie?

JohnGinnane:
I still think it's very unfair for them. Antag checking will be abused very fast

asskek:
It's supposed to be powerful against them. Cult is OP. You can summon Narnar in 7 minutes.

ArcLumin:
Can confirm, cult without research is op as fuck

and we already have antag checking for cultists. It's called holy water.

ArcLumin:
for those who don't know, cultists very quickly start spouting gibberish if they're given holy water, while normal people don't. Giving them more holy water makes them deculted

doobiedoo23:
antag checks that are this easy are fucking stupid asskek, you should know this

asskek:
right, doobie, but at the same time you're also doing a 75u/75u water+potassium explosion. you're not going to do this in the middle of the hall if you don't want to cop a ban. once i revert it back to being 75u/75u, you'll need to use two large beakers in a grenade or a bluespace beaker to mix it, as well, further balancing it.

ghost:
This is still bad.
Cult nerfs need to be shit like re-adding cult research.
Re-add cult research.

asskek:
how is this bad? this allows for people to fight against cult late-game. normally fighting late-game cult is impossible.

ghost:
This allows people to fight cult 24/7.
Late game cult is an issue when sec is bad and because cult's power is directly proportional to the amount of cultists rather than a factor of cultist plus known words.

KittyKathy:
One way i can think to abuse this is that the explosion range is lower than the range it sets cultist on fire so if you capture a potential cultist you can indirectly bomb them, as in far enough the bomb damage will not hurt them by close enough it sets them on fire if they are a cultist.

But that would be stupid to do because if you captured a suspected cultist and have holy water to waste might as well just feed it to them and waste less time on them.

Also you should make the distance for the revenant to get affected scale with amount like it does for cultist too.

Quick and easy how to make holy bombs
step 1: gather 2 large beakers, grenade assembly, printed bible from the library
step2:grind the bible for 100u holy water and fill the other with potassium
step3:complete grenade then run into a crowd of cultist and blow up bonus points if it is a Narsie summoning rune

> sleep(20)

I thought using sleep was bad and causes issues so it should be replaced with spawn or does it not matter?

KittyKathy:
Oh should this not also screw vampires?

ghost:
Pretty sure grinding the bible for holy water was removed. There is no way to make holy water without a chaplain.

ArcLumin:
Still works, did it yesterday

ArcLumin:
And yes kathy is right sleep in chems is a bad fuckin idea

ArcLumin:
unless you make a sleep chem that just calls sleep because that would be fucking awesome

SpankMasterson:
+1 may be a bit OP but i like the idea

asskek:
What's the problem with calling sleep?

JohnGinnane:
When you do sleep(n) the code waits for it to finish before resuming the proc you're in, while spawn(n) runs the code in the block n deciseconds later but the code immediately after runs as usual:

http://www.byond.com/docs/ref/info.html#/proc/sleep
http://www.byond.com/docs/ref/info.html#/proc/spawn

Basically sleep is bad for chem effects because there's lots of other stuff that happens and you shouldn't be waiting

PR #3363, opened 2017-01-17
Buffs Penned Nerf Darts

ArcLumin:
Adding a pen to a nerf dart now boosts the damage of the dart from 5 to 11

### DIFF
--- a/code/modules/projectiles/ammunition/ammo_casings.dm
+++ b/code/modules/projectiles/ammunition/ammo_casings.dm
@@ -313,7 +313,7 @@
 		if(!user.unEquip(A))
 			return
 		A.loc = BB
-		BB.damage = 5
+		BB.damage += 10
 		user << "<span class='notice'>You insert [A] into [src].</span>"
 	return
 


ghost:
Heh no, I don't want this to become an actual cost effective strat to buy foam force and load pens.

blobbernaut:
it does 11 damage a shot, why not, that's not that much

ArcLumin:
You need to also add in the pen after every shot, as it doesn't keep the pen inside of it

JohnMcLaime:
Dont be dumb, it would take a lot of preparation time to get proper ammount of ammunition upgraded and it would still require a lot of hits to kill someone. Better buy ebow or emag and steal stunguns.
Fucking lol. You would rather have a shitty 1 hit kill combat but not this?
Come on.

ArcLumin:
And it doesn't boost the damage very high at all. Good luck even getting 10 pens you need to get a crit with this, screwdrivering 10 foam darts, addings 10 pens to it, adding all 10 foamdarts back in, and then making all shots hit. If you fuck up you gotta prep em again

MethMedic:
Would be good as a finishing weapon

ArcLumin:
"you just got killed by a nerf dart nigga"

asskek:
I like this, allows for cool Owl gimmicks

ghost:
haha yes
go ahead
+1

KittyKathy:
Hmm where can you get a ton of pens....
Also it would be cool if you could insert a sleepy pen into the foam dart for a improvised syringe gun which is reusable if you miss.
Can you get the pen back from the foam dart i am not sure you can.  Sucks if you accidentally used your energy pen to make one. Why can't this fire activated energy pens!

ArcLumin:
it can. However it just drops the activated pen in front of the target. Any pen inside the dart drops after the shot, not affecting the shot's damage directly.

asskek:
dude holy shit!! imagine firing activated energy pens like a makeshift laser rifle. 
kitty's idea is genius

LuisReyesX:
You can already put sleepy pens inside normal foam darts and it gives them the effect, so this is actually a good balanced change

KittyKathy:
Why does it delete the foam dart it uses and drops the pen. Would be a waste to use riot darts because it does not do stamina damage and it deletes it when used.
Also why is it dropping the pen in front of them instead of making it hit them it would be useless right now to use any pen other than a ordinary one.
I tried in a test server unless it is RNG dependent this is not a ranged sleepy pen gun disappointing.

JohnMcLaime:
Make pens embed?

ArcLumin:
Yeah just use an ordinary pen. Would be neat. But I'd need to heavily change the current code to make it so that pens do an attack by when in the nerf dart. This is just a small damage tweak

Pyko1:
Should really make it drop penned nerf darts, with a chance of embed.
And adjust the damage accordingly.

ArcLumin:
Need to alter how this code is done considerably, but I'll take a look at it. Might have to make it a tad snowflakey. Foamdarts already are a bit snowflakey so it might snowball. But yeah, prolly something like adding the force of the pen to the damage of the projectile and then giving it additional properties based on the path of the pen (i.e. high damage for edagger, giving the sleepy pen effect for the sleepy pen)

SpankMasterson:
+1 mcbaw experiences loss of opinion????

ArcLumin:
Code review needed then it's good to merge.

Pyko1:
Code seems fine, even tho it might be better using set damage instead of adding on to other, as someone might just do herp derp buff without knowing the whole deal.

PR #3366, opened 2017-01-17
Buffs Experimentor

ArcLumin:
Original relics, that is, not the chaplain kind.

Gives initial BTC to the experimentor (chance for good shit to happen, naturally at 0 so good things never happen until you boost it which can take 30 minutes for 1%)
Slightly reduces reset time
Adds BTC to the better options so it doesn't take half the round for the good shit to happen
Removes crit reaction req
Reduces powerdrain on one failure (this would make the room basically unusable before)
Reduces RNG on relic CD
Pet spawning relic doesn't QD on failure, but the failure chance has been increased from 60% to 80%. Cooldown still in effect.
Explosive relic is reusable (so long as you survive, which you prolly won't, even though it's a small explosion)

Remember, the explosive needs to be in your hands to work; it's only usable as an ahallu akbar

### DIFF
--- a/code/modules/research/experimentor.dm
+++ b/code/modules/research/experimentor.dm
@@ -26,8 +26,8 @@
 	var/mob/trackedIan
 	var/mob/trackedRuntime
 	var/obj/item/loaded_item = null
-	var/badThingCoeff = 0
-	var/resetTime = 15
+	var/badThingCoeff = 10
+	var/resetTime = 10
 	var/cloneMode = FALSE
 	var/cloneCount = 0
 	var/list/item_reactions = list()
@@ -256,7 +256,7 @@
 	////////////////////////////////////////////////////////////////////////////////////////////////
 	if(exp == SCANTYPE_POKE)
 		visible_message("[src] prods at [exp_on] with mechanical arms.")
-		if(prob(EFFECT_PROB_LOW) && criticalReaction)
+		if(prob(EFFECT_PROB_LOW+badThingCoeff))
 			visible_message("[exp_on] is gripped in just the right way, enhancing its focus.")
 			badThingCoeff++
 		if(prob(EFFECT_PROB_VERYLOW-badThingCoeff))
@@ -280,7 +280,7 @@
 	////////////////////////////////////////////////////////////////////////////////////////////////
 	if(exp == SCANTYPE_IRRADIATE)
 		visible_message("<span class='danger'>[src] reflects radioactive rays at [exp_on]!</span>")
-		if(prob(EFFECT_PROB_LOW) && criticalReaction)
+		if(prob(EFFECT_PROB_LOW+badThingCoeff))
 			visible_message("[exp_on] has activated an unknown subroutine!")
 			cloneMode = TRUE
 			cloneCount = badThingCoeff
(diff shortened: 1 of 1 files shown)


KayeArray:
explosive + bombsuit = fun

ArcLumin:
It's still a pretty small explosion, but yeah, that's why I wanted it.

Spacedong:
how about you fucking explain your acronyms like BTC or fucking CD goddamn

KayeArray:
how abut u git smart

ArcLumin:
BTC is BadThingCoeff
CD is cooldown

Spacedong:
headmins aren't coders and you need their +1's fag, explain it more simple

ArcLumin:
If you can't understand this then you've never used experimentor

blobbernaut:
:what:

Kierany9:
> +//			qdel(src) //Comment this line to produce a light grenade (the bomb that keeps on exploding when used)!! 

# ARC YOU MADMAN!!

Spacedong:
wtf

ArcLumin:
das me

ghost:
heck yeah experimentor is otherwise kuso
+1

JohnMcLaime:
Experimentor is fucking garbage, some rp kiddo thought it would be a good idea to give it such limited shitty use. Also the shitcode of it which lets you dupe items (?)

KittyKathy:
Hey uh so this light grenade can be used with TK from range to allah snackbar chumps?

JohnMcLaime:
That would be hilarious.

ArcLumin:
They're slow, (wind up time) have an obvious message, very small explosion, and a cooldown. A potass/water nade collection is easier to get with a similar outcome. And no. There's a check after the windup. The user and the item need to be in the same place

SpankMasterson:
+1

KittyKathy:
I'm having a hard time understanding the exploding relic code this is it as it is now.
`explosion(user.loc, -1, rand(1,5), rand(1,5), rand(1,5), rand(1,5), flame_range = 2)`
This is the explosion proc as it is now.
`/proc/explosion(turf/epicenter, devastation_range, heavy_impact_range, light_impact_range, flash_range, adminlog = 1, ignorecap = 0, flame_range = 0 ,silent = 0)`

So the first argument is the location the second is devastation range as in gib range set to -1 so it does not gib or destroy items always. Third argument is heavy impact range which is random from 1 to 5 same for the light impact and flash range. Why is the sixth argument random from 1 to 5 when that should be admin log which should be 0 for no log or 1 to log it?

Only makes sense if it was originally.
`explosion(user.loc, rand(1,5), rand(1,5), rand(1,5), rand(1,5), flame_range = 2)`
Yes a random devastation range is probably op and would destroy the object that caused the explosion too so no light grenade. 

Also this is a small explosion in the sense of range but in terms of power it is random but capable of doing more damage than a 100 water + 100 potassium grenade which is I think Devastation: 0 Heavy: 2 Light: 5. No random from 1 to 5 flame range why.
maybe rewriting it to this would be better with probably lower range than 5.
`explosion(user.loc, -1, rand(1,5), rand(1,5), rand(1,5), flame_range = rand(1,5))`

Another thing is the excessive admin warning like it calls warn_admins every time a relic is used even the useless/harmless ones like Rapid duplicator, Corgi Cannon, Teleport and the one to make a cleaner grenade effect is called Smoke???

KittyKathy:
Oh wait only certain ones are priority which they actually see real time I think.

ArcLumin:
Yeah, experimenter code is weird as fuck.

ArcLumin:
Code review then this is good to merge

Pyko1:
I don't see anything bad codewise.

PR #3386, opened 2017-01-21
Fixes farts

tortellinitony:


gumner:
make so that scrubbers can succ it and it will be balanced

tortellinitony:
they already can

gumner:
ooh

ArcLumin:
Protip; farts spawn at a fairly colder temperature than normal so people tend to take minor temperature damage

PopNotes:
copypasting my idea from the shitty spacedong post that sucked

Rework, rework, rework.
Rework instead of fucking removing everything like a shitler.
Make fart gas dissipate really quickly and decay much faster, problem solved and you can still choke someone with your ass blasts so long as you stay on the same tile.

PopNotes:
I view the only real (aside from bugs/lack of visibility on analyzers) issue to be the fact that assgas stays in the air for fucking centuries when it should spread out super quick and decay pretty fast too, think the way a real poot would

